PR #20, opened 2021-11-03
feat: dids readme to the docs page

Dudleyneedham:
## fixes https://github.com/KILTprotocol/ticket/issues/1678

Moving the readme of DID to the docs page

## How to test:

yarn build
yarn start

## Checklist:


ntn-x2 (changes requested):
I would put the section about CTypes right after the DID section, then the sections about claiming, attestations, requesting terms, verification, and delegations, in this order. I would also perhaps like to have one or two lines acting as glue between the different sections, if we change the order between them.

Dudleyneedham:
I have changed the order but went with requesting terms before the attestation.

Dudleyneedham:
I wont add these sentences in this PR. I do agree this should be done, but outside the scope of the ticket and should be created as a new ticket.

ntn-x2 (approved):
Are the integration tests supposed to pass?

Dudleyneedham:
At the moment, they are failing and have been failing for some time. They do require updating.

PR #25, opened 2021-11-11
docs: update collator docusaurus

ntn-x2:
This PR fixes KILTProtocol/ticket#189.

## Changes

I have made few changes, especially about:

- fixing outdated information
- making the whole documentation a bit more formal by removing most of the you and we pronouns
- Re-arranging some paragraphs around for improved clarity


docs/chain/collator.md
```diff
@@ -1,136 +1,116 @@
...
+Since the WASM runtime logic is part of the blockchain state itself and hence represents the single source of truth, all collators should execute the WASM version of the runtime logic.
 
-#### Select which runtime and chain to use
+#### Peregrine and Spiritnet chain specs
 
-The `--chain` parameter decides which blockchain we want to execute.
+The `--chain` parameter decides which blockchain the KILT collator will join.
 We need to pass this parameter to the parachain and the relaychain, since both chains are separate blockchains.
-The parachain accepts an additional runtime parameter that selects which runtime is executed and decides which execution environment is used.
+The KILT parachain accepts an additional runtime parameter to select which runtime is executed.
```
weichweich:
```suggestion
The KILT parachain accepts an additional runtime parameter to select which runtime environment is used to executed the WASM runtime.
```

docs/chain/collator.md
```diff
@@ -1,136 +1,116 @@
...
 
-### Acquire the executable
+### Run a Peregrine collator
 
 <Tabs
   groupId="exec-strategy"
   defaultValue="Docker"
   values={[
-    {label: 'Native', value: 'Native'},
+    {label: 'Binary', value: 'Binary'},
```
weichweich, line 112: The Docker option doesn't use a binary? I prefer native since that's the difference in my opinion.

docs/chain/collator.md
```diff
@@ -1,136 +1,116 @@
...
-We recommend following the [repository](https://github.com/KILTprotocol/mashnet-node).
-Below is the command to connect as a Collator in the root directory of the mashnet-node repository.
+We recommend following the instructions in the [KILT chain repository](https://github.com/KILTprotocol/mashnet-node).
+Below is the command to build the KILT collator executable for the Peregrine network. The command must be run from the root directory of the repository after it has been cloned.
 
 :::info
 
-Do not use master branch to compile the build.
-Either use `develop` or the following tag `1.1.1`.
+We discourage to use the `develop` branch to build the executable. Instead, the last released version should be used.
```
weichweich: Not sure about that. It only matters if they use the native runtime.

docs/chain/collator.md
```diff
@@ -210,53 +192,50 @@ docker run -p 127.0.0.1:9933:9933 -v ~/data:/data \
...
-## Sync Data
-
-Before a Collator can author blocks, the node needs to fully sync up with the blockchain.
-Depending on which chain and the size of it may take a number of minutes to several hours maybe even a day.
+## Sync the Blockchain State
 
-More details can be found at the [Polkadot network](https://wiki.polkadot.network/docs/maintain-guides-how-to-validate-kusama#synchronize-chain-data)
+Before a collator can author blocks, the node needs to fully sync up with both the parachain and the relaychain.
+Depending on the size of the blockchain states, it may take a number of minutes to several hours maybe even days for the node to catch up.
+More details can be found on the [Polkadot network docs](https://wiki.polkadot.network/docs/maintain-guides-how-to-validate-kusama#synchronize-chain-data).
```
weichweich: Don't give them false hopes.

```suggestion
Depending on the size of the blockchain states, it may take a number of hours to a couple of days for the node to catch up.
More details can be found on the [Polkadot network docs](https://wiki.polkadot.network/docs/maintain-guides-how-to-validate-kusama#synchronize-chain-data).
```

docs/chain/collator.md
```diff
@@ -210,53 +192,50 @@ docker run -p 127.0.0.1:9933:9933 -v ~/data:/data \
...
-A Collator can call a RPC to check whether the account has session keys with the following calls `hasKey` and `hasSessionKeys` in `author`.
+Once the chain has fully synced, the collator needs to set sessions keys to begin collating on the network.
+To check whether the account has already some session keys set, the RPC `hasKey(publicKey, keyType)` and `hasSessionKeys(sessionKeys)` in the `author` can be called.
 
 ![](/img/chain/author-hasKey.png)
 
 :::note
 
-The Session keys associate a collator node with an account on KILT
+The session keys associate a collator node with an account on KILT.
```
weichweich:
```suggestion
The session keys associate a collator node with an account on the KILT blockchain.
```

docs/chain/collator.md
```diff
@@ -323,176 +297,175 @@ Secret phrase `very secure private key you should not use the example private ke
...
 
 </TabItem>
 </Tabs>
 
-The collator has a session key by following the previous steps.
-The collator must then make an extrinsic by signing and submitting a transaction before the collator can begin collating on the network.
-Visit the [polkadot apps and connect to the peregrine network](https://polkadot.js.org/apps/?rpc=wss%3A%2F%2Fperegrine.kilt.io#/explorer) (websocket address: `wss://peregrine.kilt.io`).
+Once a new session key is generated, the collator must then link that key to its own account in order to receive rewards for producing new blocks.
+This operation is performed by submitting a signed extrinsic to the blockchain.
+Visit the [PolkadotJS Apps and connect to the peregrine network](https://polkadot.js.org/apps/?rpc=wss%3A%2F%2Fperegrine.kilt.io#/explorer) (websocket address: `wss://peregrine.kilt.io`).
```
weichweich: peregrine or spiritnet

weichweich (commented):
Will finish tomorrow. only got through half of it and it's already 1000 times better.

docs/chain/collator.md
ntn-x2: For me it was confusing as we are not running the native runtime, bus still running the wasm version. Or maybe I got it wrong and Native indicates that the native runtime version is used? Maybe binary is not the best word choice, but I think neither is native if we then use `--execution=wasm`.

docs/chain/collator.md
ntn-x2: I did not change that bit, I just copied it over from further down the text 😂

docs/chain/collator.md
ntn-x2: Yeah that's the thing. The whole initial part is about peregrine, and then only mentions spiritnet in its section. I would argue that it might be best to actually just put a small disclaimer in the beginning saying that the operations should be first tested out on peregrine, and perhaps provide the WS endpoint for that, but then only use spiritnet in the step-by-step guide. What do you think?

docs/chain/collator.md
weichweich: Ok yes didn't see that it's overloaded with the native runtime vs wasm runtime. good point. Leave it like it is...

docs/chain/collator.md
weichweich: would be cool to have a switch in the beginning of the doc and then we just show either peregrine or spiritnet data where appropriate. Like a Tab over the whole document. :D

docs/chain/collator.md
weichweich: would be cool to have a switch in the beginning of the doc and then we just show either peregrine or spiritnet data where appropriate. Like a Tab over the whole document. :D

docs/chain/collator.md
```diff
@@ -323,176 +297,175 @@ Secret phrase `very secure private key you should not use the example private ke
...
 :::
 
-## Add collator
+## Join the collator candidates pool
 
-:::info
+:::warning
 
-It is recommended to set the session keys before doing the following steps.
+These steps should be followed only once a collator node has successfully linked a session key to its address by following the steps above.
```
weichweich: i would add:

> And the node is fully synchronized with the blockchain, so that it can build blocks.

docs/chain/collator.md
```diff
@@ -323,176 +297,175 @@ Secret phrase `very secure private key you should not use the example private ke
...
-The collator must make an extrinsic call from the `parachainStaking` pallet with the method `joinCandidates` with the desired stake entering the `candidatePool`.
+The collator must call an extrinsic from the `parachainStaking -> joinCandidates(stake)` with the desired stake to join the candidate pool.
 
 `Developer -> Extrinsics -> Submission`
 
-1. Using the selected account: Select the Collator Account 
-2. Submit the following extrinsic: parachainStaking -> joinCandidates(stake)
-3. Insert the staked KILT amount for the collator 
+1. Select the collator's KILT address as the extrinsic submitter (the *using the selected account* field)
+2. Submit the following extrinsic: `parachainStaking -> joinCandidates(stake)`
```
weichweich: We first select the extrinsic and set parameters!

```suggestion
2. Pick the following extrinsic: `parachainStaking -> joinCandidates(stake)`
```

docs/chain/collator.md
```diff
@@ -323,176 +297,175 @@ Secret phrase `very secure private key you should not use the example private ke
...
+3. Insert the staked KILT amount for the collator
+
```
weichweich, line 409:
```suggestion
4. Sign and send the extrinsic

```

docs/chain/collator.md
```diff
@@ -323,176 +297,175 @@ Secret phrase `very secure private key you should not use the example private ke
...
+:::info
+
+A recent change in the blockchain metadata resulted in a change in the UI regarding how balances are shown. In the current version of PolkadotJS Apps, specifying 1 KILT requires adding 15 trailing `0`s. So, for instance, 1 KILT needs to be written as `1000000000000000`, while 10,000 KILT would be written as `10000000000000000000`.
```
weichweich: A sentence per line! 😋 

```suggestion
A recent change in the blockchain metadata resulted in a change in the UI regarding how balances are shown.
In the current version of PolkadotJS Apps, specifying 1 KILT requires adding 15 trailing `0`s.
So, for instance, 1 KILT needs to be written as `1000000000000000`, while 10,000 KILT would be written as `10000000000000000000`.
```

docs/chain/collator.md
```diff
@@ -323,176 +297,175 @@ Secret phrase `very secure private key you should not use the example private ke
...
+
+:::
 
 ![](/img/chain/parachainStaking-joinCandidates.png)
 
-A collator can check the current `SelectedCandidates` to see the position and required staked amount to join the network. 
+A collator can check the current top candidates to see the position and required staked amount to join the network.
```
weichweich: Maybe we should start to distinguish between collator candidates and active collators.

```suggestion
A collator candidate can check the current top candidates to see his position and required staked amount to become an active collator.
The 75 collator candidates with the highest stake are shown in the top candidate list.
```

docs/chain/collator.md
```diff
@@ -323,176 +297,175 @@ Secret phrase `very secure private key you should not use the example private ke
...
 
-1. Selected state query: `parachainStaking -> selectedCandidates(): Vec<AccountId>`
-2. Add the item with the "+" on the right side
+1. Selected state query: `parachainStaking -> topCandidates(): ParachainStakingSetOrderedSet`
+2. Execute the query by pressing the "+" button on the right side
 
-If the collator has enough self-stake and delegator stake they can be selected to collate.
-Once the collator has been chosen, the collator will be added to the `SelectedCandidate` pool.
+If the collator has enough self-stake and delegator stake it will be selected to collate.
+Once the collator has been chosen, it will be part of the `topCandidates()` set.
```
weichweich:
```suggestion
A collator has been chosen, when he is part of the first 17 candidates in the `topCandidates()` set (16 for peregrine).
```

docs/chain/collator.md
```diff
@@ -323,176 +297,175 @@ Secret phrase `very secure private key you should not use the example private ke
...
+The corresponding extrinsics for these operations are `parachainStaking -> candidateStakeMore(more)` and `parachainStaking -> candidateStakeLess(less)`.
 
 ![](/img/chain/parachainStaking.png)
 
 ## Monitoring
 
-It would be ideal if monitoring is done from a different host.
-However in cases of resource constraints it would be better to run these monitoring tools on the same machine where your collator is running than not having.
-You can either run grafana by your own or subscribe free [grafana cloud ](https://grafana.com/products/cloud/) option. 
+It would be ideal if the host being monitored is not the host monitoring, a.k.a., if the monitoring process does not run on the same host as the collator process.
```
weichweich: I think the comma is after i.e. [Had to duckduckgo that](https://www.wikihow.com/Use-i.e.-in-a-Sentence)

```suggestion
It would be ideal if the host being monitored is not the host monitoring i.e., the monitoring process does not run on the same host as the collator process.
```

docs/chain/collator.md
```diff
@@ -323,176 +297,175 @@ Secret phrase `very secure private key you should not use the example private ke
...
+:::info
+
+For cloud-based solutions, the prometheus process must be publicly accessible, e.g., via a reverse proxy.
+
+:::
+
+### Prerequisites
+Install the latest version of docker-compose from the [official docker-compose installation guide](https://docs.docker.com/compose/install/), then:
+
+1. Clone or download the monitoring template from the [KILT chain repo](https://github.com/KILTprotocol/docs)
```
weichweich:
```suggestion
1. Clone [the repository](https://github.com/KILTprotocol/docs) or download [the monitoring template](https://github.com/KILTprotocol/docs/tree/master/collator).
```

docs/chain/collator.md
```diff
@@ -323,176 +297,175 @@ Secret phrase `very secure private key you should not use the example private ke
...
+2. Change directory to the above with ```cd docs/collator```
+3. Edit the `.env` file and insert grafana admin and password
+4. Depending on the installation type either:
+  - run `docker-compose up -d` to install only prometheus and node exporter or 
+  - run `docker-compose up --profile grafana -d` to install prometheus through grafana or
```
weichweich:
```suggestion
  - run `docker-compose up --profile grafana -d` to install node exporter, prometheus and grafana or
```

docs/chain/collator.md
```diff
@@ -323,176 +297,175 @@ Secret phrase `very secure private key you should not use the example private ke
...
+  - run `docker-compose --profile collator --profile grafana up -d` to install prometheus **and** grafana with the collator
```
weichweich:
```suggestion
  - run `docker-compose --profile collator --profile grafana up -d` to install node exporter, prometheus, grafana and **a collator node**
```

docs/chain/collator.md
```diff
@@ -323,176 +297,175 @@ Secret phrase `very secure private key you should not use the example private ke
...
 
-NB: if you want to use cloud option you need to expose prometheus to outside access via reverse proxying with webserver.
-The following Monitoring stack is recommended.
+:::info
+
+For cloud-based solutions, the prometheus process must be publicly accessible, e.g., via a reverse proxy.
+
+:::
+
+### Prerequisites
```
weichweich:
```suggestion
### Installation
```

docs/chain/collator.md
```diff
@@ -323,176 +297,175 @@ Secret phrase `very secure private key you should not use the example private ke
...
+The monitoring infrastructure can either be run as a local grafana cluster or as a [cloud-based solution](https://grafana.com/products/cloud/).
 
-NB: if you want to use cloud option you need to expose prometheus to outside access via reverse proxying with webserver.
-The following Monitoring stack is recommended.
+:::info
+
+For cloud-based solutions, the prometheus process must be publicly accessible, e.g., via a reverse proxy.
+
+:::
+
```
weichweich, line 453: A section that describes what will be installed would be good?

```suggestion

# What will be installed

The docker compose starts up to four containers:

* **Node Exporter**: The node exporter provides metrics from your host, like CPU, memory and storage usage, network traffic statistics etc.
* **prometheus**: Prometheus collects and stores metrics that it collected from the node exporter and your blockchain node
* **grafana (optional)**: Grafana is a website which shows the metrics and can send alerts on certain conditions
* **collator (optional)**: A KILT spiritnet node that is configured to as a collator
```

docs/chain/collator.md
```diff
@@ -323,176 +297,175 @@ Secret phrase `very secure private key you should not use the example private ke
...
+The configuration can be checked by visiting `https://localhost:3000` and authenticating with the username and password set in `.env` at step 3.
+
+### Configuring Alert Notification Channel
+Choose any of the supported notification channels and follow the [Grafana documentation](https://grafana.com/docs/grafana/latest/alerting/old-alerting/notifications/) to receive alerts and notifications.
+
+Overall, for monitoring we recommend the following stack:
 - Prometheus
 - Grafana
 - Node exporter
 - Slack
```
weichweich: I wouldn't recommend slack specifically.

```suggestion
```

docs/chain/collator.md
```diff
@@ -574,21 +547,22 @@ docker run -p 127.0.0.1:9933:9933 -v ~/data:/data \
...
+The same process that includes generating and registering sessions keys and joining the collator candidate pool must be re-executed for Spiritnet.
 
-- running "cargo update -p environmental" could fix something
-- Using the commands create a `/Data/` directory the folder could possibly be owned by the root user. Switching the ownership of the folder with `chown [options] user:group filename`. Solution found [here](https://serverfault.com/questions/106713/how-to-change-ownership-of-a-folder-using-terminal)
+:::
 
 ## Troubleshooting
 
-There are a few things that you can check to make sure everything is setup correctly.
+There are a few things that can checked to make sure everything is setup correctly.
```
weichweich: i would keep this note:

```
Check that your node is fully synced with the relaychain & parachain. (best and finalised block number is equal to the one shown in the polkadot apps & on subscan)
```

docs/chain/collator.md
```diff
@@ -574,21 +547,22 @@ docker run -p 127.0.0.1:9933:9933 -v ~/data:/data \
...
+
+If, from any network explorer, e.g., the one offered by PolkadotJS Apps, the collator's account is shown next to some of the blocks, then the collator is correctly producing blocks and getting rewarded for it.
+If the logs print the message that starts with a :gift: emoji it indicates that the collator setup is correct but that the blocks produced are not included by the relaychain. This typically signals some issues about the node hardware or connectivity.
```
weichweich, line 565: This is only true if the first step failed. That's the reason i added numbers. It's like if n-1 failed check n.

```suggestion
If the logs print the message that starts with a :gift: emoji it indicates that the collator setup is correct but that the blocks produced are not included by the relaychain. This typically signals some issues about the node hardware or connectivity.
```

docs/chain/collator.md
ntn-x2: So the question is: what message to we want to convey here? That `master` should not be used, like it was before, or that it is, on the contrary, preferable even though not required?

docs/chain/collator.md
ntn-x2: That would mean duplicating the same content twice, right? I would prefer to avoid that. But I will think about a better way to do that in a separate commit after addressing your comments and will ask you to review that.

docs/chain/collator.md
weichweich: Ah right! I didn't notice the part about master. Yes so they could use both. But we might want to tell them to only use master.

docs/chain/collator.md
ntn-x2: I would refrain from putting actual numbers, as those need to be updated. Plus, we have mentioned few lines above that collators can check the minimum stake required to make in the active set, so I am not sure why we want to repeat that here.

docs/chain/collator.md
ntn-x2: Yeah wanted to be a bit more informal here, but ok 🧐 It goes both before and after tho.

docs/chain/collator.md
ntn-x2: It is, just a few points below, precisely point 3. I just switched the points around as to first suggest to check if everything is good, and if not, rule out potential causes before trying to upgrade the hardware. I think it is a more logical flow to follow.

docs/chain/collator.md
ntn-x2: So after which point you would include this?

docs/chain/collator.md
ntn-x2: So `master` or the latest release?

docs/chain/collator.md
weichweich: It would be cool to have switch on the top that switches between peregrine & spiritnet and that just replaces the values everywhere. 
I couldn't find something like that... But yes. Definitely don't put a tab for the whole doc! :D

docs/chain/collator.md
weichweich: yes that's also fine. But the TopCandidates are candidates and only the first N candidates will become collators. It reads like everyone in the list is a collator

docs/chain/collator.md
weichweich: ok!

docs/chain/collator.md
weichweich: i think the order is fine. it wasn't clear for me that this only matters when the first point was false.
It sounds like having the 🎁 in logs is something bad. Like it means that the block wasn't submitted.

docs/chain/collator.md
ntn-x2: I would then remove this line altogether, as I think it does not bring anything new on the table.

docs/chain/collator.md
weichweich: master i would say

weichweich (approved):

docs/chain/collator.md
ntn-x2: Let me know what you think https://github.com/KILTprotocol/docs/pull/25/commits/fc55280db6c574298a4fe65caf4e65087efe346e

weichweich: love it!

PR #35, opened 2021-12-06
feat: how to remove a vote

weichweich:
We got the question a lot. Let's have a guide


Dudleyneedham:
Do you think this should be a full voting page? It seems it would be better to have the workflow of voting from voting, updating and removing.

Dudleyneedham:
Additionally, we could advertise where people can view the governance motions/proposals/etc.

weichweich:
Yes I think this guide is one of many to follow ;)

PR #44, opened 2022-01-25
Split Workshop into claimer, verifier, attester roles

paulpomerleau:
## Modifies code structure 
Splits the code into Claimer, Verifier, Attester so that the user has real world workflows to draw from. Adds .env variables as well to help promote good practice. Updated docs to match

## How to test:
make sure the `.env` file is empty and run `node code-examples/workshop/index` 
  
# Some things left to do on this branch before merging
- [x] Go through and simplify the text
- [ ] Setup index.ts with proper tests
- [ ] Provide a link to Peregrine faucet
- [x] Provide a link to workshop code


weichweich:
very nice! Maybe we could add separate folders for attester & claimer?

paulpomerleau:
the attester, claimer and verifier have separate folders in this update... although there is a main entry file. this is just to avoid the dev having to employ comms between node processes. in the real world it would be done over the wire.

paulpomerleau:
> very nice! Maybe we could add separate folders for attester & claimer?

ahh I see you meant for the markdown, yes good idea

weichweich:
> > very nice! Maybe we could add separate folders for attester & claimer?
> 
> ahh I see you meant for the markdown, yes good idea

Yeah I first try to make everything simpler and then i want to move the workshop source files to it's own repository. Not sure how I will prevent code duplication. Maybe I try git submodules or something. Lets see. 😓

paulpomerleau:
We could just strip comments, compress and provide a download link? It's
just for accessibility.. doesn't necessarily need to be a repo 🤷.

PR #51, opened 2022-02-16
tests: fix workshop tests & move core features

weichweich:
* run workshop without cleaning up afterwards.
* move core_features into a separate folder

## Checklist:


.github/workflows/test.yml
```diff
@@ -17,25 +17,30 @@ jobs:
     steps:
       - uses: actions/checkout@v1
 
+      - name: Use Node.js 16.x
```
Dudleyneedham, line 20: we should use node 14

Dudleyneedham (approved):
LGTM! One minor comment

.github/workflows/test.yml
weichweich: the SDK uses either 14 or 16?

PR #57, opened 2022-03-02
deps: update SDK to v0.26

ntn-x2:
## Update SDK to v0.26

This PR is part of https://github.com/KILTprotocol/ticket/issues/1840 by updating the workshop to use v0.26 of the SDK. Specifically, it updates the `core_features` and `workshop` folders to use the new SDK version.

### What it does not do

There are other places which contain outdates JS snippets, which are never checked and which have become obsolete. I have removed most of the JS snippets, since they were outdated anyway, but a separate PR (which would be tracked by https://github.com/KILTprotocol/ticket/issues/1863) should take care of them, by using something similar to how the code in `core_features` and `workshop` is checked.

## Checklist

- [x] Get reviewed
- [ ] Update to latest release when out
- [ ] Test with new latest release
- [ ] Merge and publish


code-examples/core_features/claiming/index.ts
```diff
@@ -4,49 +4,46 @@ import {
   DemoKeystore,
   SigningAlgorithms,
 } from '@kiltprotocol/did'
+import { VerificationKeyType } from '@kiltprotocol/sdk-js'
 
 import { main as main1 } from './1_claiming'
 import { main as main2 } from './2_claiming'
 import { main as main3 } from './3_claiming'
 
-const SEED_ENV = 'FAUCET_SEED'
```
ntn-x2, line 12: Was not used anywhere.

code-examples/core_features/did/9_did.ts
```diff
@@ -0,0 +1,76 @@
...
+      name: {
+        type: 'string',
+      },
+      age: {
+        type: 'integer',
+      },
+    },
+    type: 'object',
+  })
+}
```
weichweich: please add a new line. 🙂

code-examples/core_features/did/3_did.ts
```diff
@@ -1,66 +1,51 @@
...
 export async function main(
   keystore: DemoKeystore,
   kiltAccount: KeyringPair,
-  resolveOn: SubscriptionPromise.ResultEvaluator,
   // Generate seed for the authentication key.
-  authenticationSeed: string
-): Promise<IDidResolvedDetails> {
-  // Initialise connection to the public KILT test network.
+  authenticationSeed: string,
+  resolveOn: SubscriptionPromise.ResultEvaluator = BlockchainUtils.IS_IN_BLOCK
```
weichweich: I like that this is now more visible to the reader. But maybe we should rather use on finalized here? In block was only used to speed up the CI.

code-examples/core_features/did/3_did.ts
```diff
@@ -1,66 +1,51 @@
...
+    })
   })
 
-  // Retrieve the newly created DID from the KILT blockchain.
-  const fullDid = await DefaultResolver.resolveDoc(did)
-
   await disconnect()
-  if (fullDid === null) {
-    throw 'Could not find DID document for the given identifier'
+  if (!fullDid) {
```
weichweich, line 47: AFAIK you should test for null with `=== null` since `!` returns true in a lot more cases.

```js
console.log(!undefined)
console.log(!"")
console.log(!0)
```

weichweich (commented):
Looks good as far as I can judge.
Not sure about the change in checking for null and a new line is missing.
I really like that you improved a few things while updating. 🐝

code-examples/core_features/did/3_did.ts
ntn-x2: We use `!` everywhere in the SDK as well. We are always dealing with objects, so the result is `false` only if the object is `null` or `undefined`, which is good enough considering it is Typescript.

code-examples/core_features/did/9_did.ts
ntn-x2: Will be caught once the linter is in! https://github.com/KILTprotocol/docs/pull/57/commits/3d518bdc8e5150315bc3403ffe1afaff2dac68c7

code-examples/core_features/did/3_did.ts
ntn-x2: https://github.com/KILTprotocol/docs/pull/57/commits/580d3baa7abf59526df2d47ede1974a469dd6361

weichweich (approved):

PR #58, opened 2022-03-02
[Fix] Adds --bootnodes on the commandline args for each bootnode

ggera:
Adds --bootnodes flag on the commandline args for each bootnode as outline by this breaking cli change here https://github.com/paritytech/polkadot/releases/tag/v0.9.17-rc4

## Checklist:

### DIFF
--- a/docs/chain/collator.md
+++ b/docs/chain/collator.md
@@ -517,16 +517,16 @@ For the sake of completeness, the bootnodes are listed below:
 
 For **spiritnet**, the parachain bootnodes are:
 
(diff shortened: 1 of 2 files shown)


docs/chain/collator.md
```diff
@@ -517,16 +517,16 @@ For the sake of completeness, the bootnodes are listed below:
 
 For **spiritnet**, the parachain bootnodes are:
```
Dudleyneedham, line 519: Should there be a note here stating "bootnodes argument is require"

Dudleyneedham (commented):
Minor question

docs/chain/collator.md
weichweich: They are not required anymore. :)

weichweich (approved):

PR #68, opened 2022-05-03
chore: rename to kilt-node

weichweich:
mashnet-node -> kilt-node

## Checklist:


tjwelde:
nice, even changed the diagram

weichweich:
> nice, even changed the diagram

That's the nice thing about SVG. It showed while i searched for mashnet-node. 😀

PR #67, opened 2022-05-02
Redo overview

weichweich:
The current overview is IMHO confusing. What does the diagram show? We don't explain a it at all or even refer to it in the text. What is the reader supposed to take out of it?

I would propose that we replace it with two sequence diagrams. One for the attestation flow and another one for the verification flow. This way the reader gains a high level overview of the two processes and it should become easier to create a mental model of what the code in the following steps does.

## Checklist:

- [ ] I have verified that the code works
- [ ] I have verified that the code is easy to understand
  - [ ] If not, I have left a well-balanced amount of inline comments
- [ ] I have [left the code in a better state](https://deviq.com/principles/boy-scout-rule)
- [ ] I have documented the changes (where applicable)

PR #76, opened 2022-05-06
fix: Mermaid darkmode

wischli:
## fixes KILTProtocol/ticket#1980

* Wraps `kilt-mermaid` class div around dark mode problematic mermaid SVGs
* Moves custom css from `style.module.css` to `custom.css` which is not required to be applied dynamically in jsx/mdx

Not sure if the wrapping is the solution we want to go with.
The problem with our MDX Mermaid plugin is that Mermaid diagrams can only be distinguished via their SVG id of the form `mermaid-id-#`.
However, the mapping of a diagram to its id is not deterministic, e.g. depending on where you start, a Mermaid SVG could have id 1 or 10. 
Currently, there exist two Mermaid diagrams  (from #67) for which we don't need to overwrite their dark mode coloring. 

fixes KILTprotocol/ticket/issues/1988

## Checklist:


ntn-x2:
@wischli this PR seems to have conflicts with the base branch. Do you mind resolving them? I could then give it a review.

weichweich (approved):

PR #82, opened 2022-05-09
docs: add DID section in concepts

ntn-x2:
## fixes KILTProtocol/ticket#1968

Copy-paste content from the previous `SDK > Core Features`.
Specifically:

- remove code snippets
- remove in-depth sections like DID creation etc.
- puts Web3 names and account linking in the same DID paragraph.

This PR does **not** remove the old content from the `SDK > Core Features` section. Does it have to?


docs/concepts/did.md
```diff
@@ -2,3 +2,82 @@
...
+A regex that would match all and only the allowed Web3 names would be the following:
+
+```
+^[a-z0-9\_\-]{3,32}$
+```
+
+In the global URI space, Web3 names are prefixed with the `w3n:` URI namespace. 
+For example, the full URI for the Web3 name `example-web3name` is `w3n:example-web3name`.
+
+### Linking accounts to a Web3 name
```
weichweich, line 79: I would love to see a simple diagram for this. Just to make it easier.

docs/concepts/did.md
```diff
@@ -2,3 +2,82 @@
...
+
+Beyond linking a Web3 name, KILT allows DID owners to link multiple accounts to a single DID.
+These accounts are not specific to the KILT blockchain but can reference any chain within the Kusama ecosystem.
```
weichweich: It's not only Kusama and it's rather limited by the used signing algorithms.

```suggestion
These accounts are not specific to the KILT blockchain but can reference any chain within the Kusama&Polkadot ecosystem that uses sr25519 or ed25519.
```

weichweich (changes requested):
I would love to see a section that compares DIDs to KILT accounts. It think this is the first thought that everyone has. What's a DID and whats a KILT account?!
We should make it clear that a did cannot have any KILTs.
Maybe we can have a nice diagram that shows that KILT accounts are on the infrastructure level and DIDs are at the application(?) level.

docs/concepts/did.md
ntn-x2: What? We do support Ecdsa.

docs/concepts/did.md
wischli: Definitely! We need to step up our diagram game. If you want to, I can provide these @ntn-x2 .

docs/concepts/did.md
wischli: Maybe we want to say `Dotsama` instead?

wischli:
> This PR does not remove the old content from the SDK > Core Features section. Does it have to?

IMO, it should remove all migrated content if there is no plan re-use it elsewhere.

docs/concepts/did.md
weichweich: > What? We do support Ecdsa.

Yeah forgot that. But the point is that it's not limited to kusama chains.

docs/concepts/did.md
ntn-x2: Phew! This cost me quite some time! Let me know what you guys think https://github.com/KILTprotocol/docs/pull/82/commits/f2ebb7497232d9fc2c92a65344f9722c4f7908ca.

docs/concepts/did.md
ntn-x2: https://github.com/KILTprotocol/docs/pull/82/commits/eaca2ed186e5dab481448978baf7acdc3bd4979e

ntn-x2:
@weichweich I added a warning at the end of the section highlighting the difference between addresses and DIDs, and that DIDs do not hold any balance -> https://github.com/KILTprotocol/docs/pull/82/commits/59dea722eca04aba8412ea60e81ab8565963aa1a

@wischli I am still not sure where the DID code snippets would go, that is why I have not removed them yet in this PR.

docs/concepts/did.md
```diff
@@ -2,3 +2,103 @@
...
 title: DIDs
 ---
+
+import ThemedImage from '@theme/ThemedImage';
+
+A KILT Decentralized Identifier (DID) is a string uniquely identifying each KILT user.
+A DID can be thought of as a container of different keys that are all under the control of the same DID subject.
+For the official W3C DID spec, please visit the [DID Core spec page](https://www.w3.org/TR/did-core/), while for the official KILT DID method specification, please visit the [KILT DID spec page](https://github.com/KILTprotocol/kilt-did-driver/blob/master/docs/did-spec/spec.md).
+
+The simplest type of KILT DID is a **light DID**, called this way because it can be generated and used offline without requiring any Internet connection (hence any connection with the KILT blockchain at all).
```
trusch, line 12:
```suggestion
The simplest type of KILT DID is a **light DID**, called this way because it can be generated and used offline without requiring any internet connection (hence any connection with the KILT blockchain at all).
```

ntn-x2: https://github.com/KILTprotocol/docs/pull/82/commits/27ff8acbe8ca3344d8a85d143427b5423803b27d

docs/concepts/did.md
```diff
@@ -2,3 +2,103 @@
...
+    light: '/img/concepts/did/did-lookup-light.png',
+    dark: '/img/concepts/did/did-lookup-dark.png',
+  }}
+/>
+
+:::caution
+While multiple accounts can be linked to a DID, it is important to notice the difference between the two.
+
+KILT *accounts* are classical blockchain accounts, that can be used to hold and send KILT tokens, as well as signing and submitting transactions.
+On the other hand, KILT *DIDs* are a L2 construct which are derived from KILT accounts, but are completely separated from them.
```
weichweich: I'm not sure if L2 is understood widely.

```suggestion
On the other hand, KILT *DIDs* are a higher level construct which are derived from KILT accounts, but are completely separated from them.
```

docs/concepts/did.md
```diff
@@ -2,3 +2,103 @@
...
+This means that **KILT DIDs cannot hold any KILT tokens**.
+DIDs are used to authorize (sign) some operations, but the resulting signature must then be submitted to the blockchain by a KILT account, which must pay for the transaction fees.
+
+Hence, even though a DID has the format `did:kilt:4rp4rcDHP71YrBNvDhcH5iRoM3YzVoQVnCZvQPwPom9bjo2e` with `4rp4rcDHP71YrBNvDhcH5iRoM3YzVoQVnCZvQPwPom9bjo2e` being a valid KILT account, there is no (immediate) relationship between the DID and this account and hence this account should not be considered in isolation as a KILT account, but only with its `did:kilt` prefix, which makes it a valid DID.
```
weichweich: This was hard to understand for me

```suggestion
Hence, even though a DID has the format `did:kilt:4rp4rcDHP71YrBNvDhcH5iRoM3YzVoQVnCZvQPwPom9bjo2e` with `4rp4rcDHP71YrBNvDhcH5iRoM3YzVoQVnCZvQPwPom9bjo2e` being a valid KILT account, there is no (immediate) relationship between the DID and this account and hence this account should not be considered in isolation as a KILT account, but only with its `did:kilt` prefix, which makes it a valid DID.
```

docs/concepts/did.md
```diff
@@ -2,3 +2,103 @@
...
+Hence, if instructed to "*send some funds to the DID by using the account after the `did:kilt` prefix*", please ignore the advice, as sending funds to a DID can result in those funds being lost without the required technical expertise.
```
weichweich: You used hence twice in a row. 
```suggestion
Therefor, if instructed to "*send some funds to the DID by using the account after the `did:kilt` prefix*", please ignore the advice, as sending funds to a DID can result in those funds being lost without the required technical expertise.
```

weichweich (approved):
few minor things.

docs/concepts/did.md
ntn-x2: https://github.com/KILTprotocol/docs/pull/82/commits/68ce5d9232ea37164d25c622112f8ae659f344b7

docs/concepts/did.md
ntn-x2: https://github.com/KILTprotocol/docs/pull/82/commits/f88c41076ae7f68ec87b4c5bef28de5cbb6051b0

docs/concepts/did.md
ntn-x2: Do you think this is clearer now? https://github.com/KILTprotocol/docs/pull/82/commits/168243953993122eb20033e147ce65acc14c828c

weichweich: Yes! very clear now. Thanks. :)

trusch (approved):
LGTM :partying_face:

tjwelde (approved):
LGTM

PR #83, opened 2022-05-09
Test docs agains new sdk rc

ggera:
## fixes KILTProtocol/ticket#1908
Whenever a new tag is add for a release candidate this jobs replaces sdk version on the docs and test it. If sucessful opens PR to update the docs to the latest SDK on failure opens an issue ticket. This should be triggered by `repository_dispatch` fron sdk branch
[Expected Issue](https://github.com/ggera/ticket/issues/5)
[Expected PR](https://github.com/ggera/docs/pull/10) 
## Checklist:

### DIFF
--- a/.github/workflows/sdk-update-test.yml
+++ b/.github/workflows/sdk-update-test.yml
@@ -61,7 +61,6 @@ jobs:
         working-directory: code_examples/core_features
         run: |
           yarn install
-
       - name: Run tests
         timeout-minutes: 60
         working-directory: code_examples/core_features
@@ -75,7 +74,6 @@ jobs:
     runs-on: ubuntu-latest
     needs: [ workshop, core_features ]
     if: success()
-
     steps:
       - uses: actions/checkout@v1
       - name: Get current package version
@@ -124,7 +122,6 @@ jobs:
     runs-on: ubuntu-latest
     needs: [ workshop, core_features ]
     if: failure()
-
     steps:
       - name: Checkout
         uses: actions/checkout@v2
--- a/.github/workflows/sdk-update-test.yml
+++ b/.github/workflows/sdk-update-test.yml
@@ -1,0 +1,147 @@
+name: test-docs-examples-2
+
+on:
+  repository_dispatch:
+    types: [ sdk-update ]
+
+jobs:
+  workshop:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v1
+
+      - name: Use Node.js 16.x
+        uses: actions/setup-node@v1
+        with:
+          node-version: 16.x
+      - name: Get current package version
+        working-directory: code_examples/workshop
+        id: kiltprotocol_sdk
+        run: echo "::set-output name=current_tag::$(grep kiltprotocol/sdk-js package.json | awk -F \" '{print $4}')"
+      - uses: jacobtomlinson/gha-find-replace@v2
+        with:
+          find: '@kiltprotocol/sdk-js": "${{ steps.kiltprotocol_sdk.outputs.current_tag }}'
+          replace: '@kiltprotocol/sdk-js": "${{ github.event.inputs.client_payload.latestTag }}'
+          include: "code_examples/**"
+          exclude: "**/*.lock"
+          regex: true
+      - name: Test style conventions
+        working-directory: code_examples/workshop
+        run: |
+          yarn install
+
+      - name: Run tests
+        timeout-minutes: 60
+        working-directory: code_examples/workshop
+        env:
+          NODE_OPTIONS: --unhandled-rejections=strict
+          FAUCET_SEED: ${{ secrets.PEREGRINE_FAUCET_SEED }}
+        run: |
+          yarn install
+          yarn run ts-node test.ts
+
+  core_features:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v1
+      - name: Get current package version
+        working-directory: code_examples/core_features
+        id: kiltprotocol_sdk
+        run: echo "::set-output name=current_tag::$(grep kiltprotocol/sdk-js package.json | awk -F \" '{print $4}')"
+      - uses: jacobtomlinson/gha-find-replace@v2
+        with:
+          find: '@kiltprotocol/sdk-js": "${{ steps.kiltprotocol_sdk.outputs.current_tag }}'
+          replace: '@kiltprotocol/sdk-js": "${{ github.event.inputs.client_payload.latestTag }}'
+          include: "code_examples/**"
+          exclude: "**/*.lock"
+          regex: true
+      - name: Test style conventions
+        working-directory: code_examples/core_features
+        run: |
+          yarn install
+
+      - name: Run tests
+        timeout-minutes: 60
+        working-directory: code_examples/core_features
+        env:
+          NODE_OPTIONS: --unhandled-rejections=strict
+          FAUCET_SEED: ${{ secrets.PEREGRINE_FAUCET_SEED }}
+        run: |
+          yarn run ts-node run_core_features.ts
+
+ create_pull_request:
+    runs-on: ubuntu-latest
+    needs: [ workshop, core_features ]
+    if: success()
+
+    steps:
+      - uses: actions/checkout@v1
+      - name: Get current package version
+        working-directory: code_examples/core_features
+        id: kiltprotocol_sdk
+        run: echo "::set-output name=current_tag::$(grep kiltprotocol/sdk-js package.json | awk -F \" '{print $4}')"
+      - uses: jacobtomlinson/gha-find-replace@v2
+        with:
+          find: '${{ steps.kiltprotocol_sdk.outputs.current_tag }}'
+          replace: '${{ github.event.client_payload.latestTag }}'
+          include: "docs/develop/03_workshop/**"
+          regex: true
+      - uses: jacobtomlinson/gha-find-replace@v2
+        with:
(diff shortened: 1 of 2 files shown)


.github/workflows/sdk-update-test.yml
```diff
@@ -0,0 +1,110 @@
+name: test-docs-examples
+
+on:
+  repository_dispatch:
+    types: [sdk-update]
+    inputs:
+      latestTag:
+        default: "0.27.0"
```
ntn-x2: Can't we just use `latest` here?

.github/workflows/sdk-update-test.yml
```diff
@@ -0,0 +1,102 @@
...
+        with:
+          node-version: 16.x
+      - uses: datamonsters/replace-action@v2
+        with:
+         files: 'code-examples/workshop/package.json'
+         replacements: '${{ github.event.client_payload.previousTag }}=${{ github.event.client_payload.latestTag }}'
+      - name: Test style conventions
+        working-directory: code-examples/workshop
+        run: |
+          yarn install --frozen
```
rflechtner: did you do a test run? I'm asking because I would expect this to fail with the `--frozen` option after having changed the package.json

.github/workflows/sdk-update-test.yml
```diff
@@ -0,0 +1,102 @@
...
+    steps:
+      - uses: actions/checkout@v1
+      - uses: datamonsters/replace-action@v2
+        with:
+          files: 'code-examples/core_features/package.json'
+          replacements: '${{ github.event.client_payload.previousTag }}=${{ github.event.client_payload.latestTag }}'
+      - name: Test style conventions
+        working-directory: code-examples/core_features
+        run: |
+          yarn install --frozen
```
rflechtner: same here

.github/workflows/sdk-update-test.yml
```diff
@@ -0,0 +1,102 @@
...
+          yarn run ts-node run_core_features.ts
+
+  create_pull_request:
+    runs-on: ubuntu-latest
+    needs: [workshop, core_features]
+    if: success()
+    steps:
+      - uses: datamonsters/replace-action@v2
+        with:
+          files: 'docs/develop/workshop/00-welcome.md,docs/develop/workshop/01-setup.md'
```
rflechtner: Do we not commit the changes to the package.json files? in that case we would have to update the lock file first, of course

.github/workflows/sdk-update-test.yml
```diff
@@ -0,0 +1,102 @@
...
+          replacements: '${{ github.event.client_payload.previousTag }}=${{ github.event.client_payload.latestTag }}'
+      - name: Create Pull Request
+        uses: peter-evans/create-pull-request@v4
+        with:
+          add-paths: |
+            code-examples/workshop/package.json
+            code-examples/core_features/package.json
```
rflechtner: the changes in the other two jobs unfortunately do not persist here I think, the jobs run completely isolated

rflechtner: also don't forget the lock file ;)

rflechtner (changes requested):
Could you double-check that the `yarn install` command actually works? I'm almost positive that must fail, because `--frozen` does not allow changes to the lock file, which must happen if we update the `package.json`

.github/workflows/sdk-update-test.yml
ggera: removed

.github/workflows/sdk-update-test.yml
ggera: Done

.github/workflows/sdk-update-test.yml
ggera: Nice catch

.github/workflows/sdk-update-test.yml
```diff
@@ -0,0 +1,148 @@
...
+           find: '@kiltprotocol/sdk-js": "${{ steps.kiltprotocol_sdk.outputs.current_tag }}'
+           replace: '@kiltprotocol/sdk-js": "${{ github.event.client_payload.latestTag }}'
+           include: "code_examples/**"
+           exclude: "**/*.lock"
+           regex: true
+       - name: Regenerate yarn lock
+         working-directory: code_examples/core_features
+         run: |
+           yarn install
+           yarn lint
```
ntn-x2: This is not needed.

.github/workflows/sdk-update-test.yml
```diff
@@ -0,0 +1,148 @@
...
+       - name: Regenerate yarn lock
+         working-directory: code_examples/workshop
+         run: |
+           yarn install
+           yarn lint
```
ntn-x2: Not needed.

.github/workflows/sdk-update-test.yml
```diff
@@ -0,0 +1,148 @@
...
+       - name: Create Pull Request
+         uses: peter-evans/create-pull-request@v4
+         with:
+           token: ${{ secrets.REPO_ACCESS_TOKEN }}
+           base: develop
+           delete-branch: true
+           title: '[Chore] Latest sdk is compatible with docs'
```
ntn-x2: Something like this:

```suggestion
           title: '[Chore] Update SDK version to latest (still compatible with the current docs)'
```

.github/workflows/sdk-update-test.yml
```diff
@@ -0,0 +1,148 @@
...
+   create_issue:
+     runs-on: ubuntu-latest
+     needs: [workshop, core_features]
+     if: failure()
+     steps:
+       - name: Checkout
+         uses: actions/checkout@v2
+       - name: Generate issue
+         run: |
+           cd .github/ISSUE_TEMPLATE
```
ntn-x2: Maybe change the working directory directly?

ntn-x2: Maybe change the working directory directly?

.github/workflows/sdk-update-test.yml
```diff
@@ -0,0 +1,148 @@
...
+         uses: peter-evans/create-issue-from-file@v3
+         with:
+           title: new sdk ${{ github.event.client_payload.latestTag }} is incompatible with docs examples
+           repository: KILTProtocol/ticket
+           token: ${{ secrets.REPO_ACCESS_TOKEN }}
+           content-filepath: .github/ISSUE_TEMPLATE/sdk-incompatible-with-docs.md
+           labels: |
+             :memo:  documentation
+             📚+release
+             🧰 sdk
```
ntn-x2: New line please 😀

.github/workflows/sdk-update-test.yml
```diff
@@ -0,0 +1,148 @@
...
+         uses: actions/checkout@v2
+       - name: Generate issue
+         run: |
+           cd .github/ISSUE_TEMPLATE
+           echo "[Either workshop or core_features test workflow](${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}) using the latest sdk ${{ github.event.client_payload.latestTag }} on docs has failed.Please fix!" >> sdk-incompatible-with-docs.md
+       - name: Create issue
+         uses: peter-evans/create-issue-from-file@v3
+         with:
+           title: new sdk ${{ github.event.client_payload.latestTag }} is incompatible with docs examples
+           repository: KILTProtocol/ticket
```
ntn-x2: Can we change the issue to be created on the docs instead? @tjwelde what do you think?

.github/workflows/test.yml
```diff
@@ -1,61 +1,61 @@
...
-      - name: Run tests
-        timeout-minutes: 60
-        working-directory: code-examples/core_features
-        env:
-          NODE_OPTIONS: --unhandled-rejections=strict
-          FAUCET_SEED: ${{ secrets.PEREGRINE_FAUCET_SEED }}
-        run: |
-          yarn install --frozen
-          yarn run ts-node run_core_features.ts
+# name: test-code-examples
```
ntn-x2: Uncomment please 😀

.github/workflows/sdk-update-test.yml
```diff
@@ -0,0 +1,148 @@
...
+          yarn lint
+
+      - name: Run tests
+        timeout-minutes: 60
+        working-directory: code_examples/core_features
+        env:
+          NODE_OPTIONS: --unhandled-rejections=strict
+          FAUCET_SEED: ${{ secrets.PEREGRINE_FAUCET_SEED }}
+        run: |
+          yarn install
```
rflechtner: that's probably unnecessary

rflechtner (approved):
LGTM after those comments are addressed!

.github/workflows/sdk-update-test.yml
ggera: I would argue issues not created under `ticket` repo will loose visibility from daily and planning

.github/workflows/sdk-update-test.yml
ggera: Changed to full path

.github/workflows/sdk-update-test.yml
ggera: Removed

.github/workflows/sdk-update-test.yml
ntn-x2: Yeah, I feel that is something we should discuss. Anyway, since broken docs are bad, perhaps I would add the high label since we are using the ticket board. Other than that, all the rest looks good 😃

ntn-x2 (approved):
Once my last outstanding comment is addressed, it looks good to me!

PR #84, opened 2022-05-10
docs: adjust DID core feature section

ntn-x2:
## fixes KILTProtocol/ticket#1974

This PR updates the old section of the DID core feature by removing update/deletion/deposit stuff, and only keeping light and full DID creation, DID migration, and DID batching.

Based on top of https://github.com/KILTprotocol/docs/pull/82.

### DIFF
--- a/docs/develop/sdk/0-core-feature/0_did.md
+++ b/docs/develop/sdk/0-core-feature/0_did.md
@@ -1,6 +1,6 @@
 ---
 id: did
-title: Kilt DIDs
+title: KILT DIDs
 ---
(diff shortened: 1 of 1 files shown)


tjwelde (approved):
LGTM

weichweich (approved):
I like the new style. The theory is covered in a different guide. This feels way closer to developers now. When you build something and you want to know how to batch things together you would come here.

Is there a reason to remove the `update` and `delete` guides?

ntn-x2:
@weichweich that is what we agree on with @tjwelde as well. He said to keep it simple to just the key operations.

PR #86, opened 2022-05-13
fix: unsafe external for docker

### LINKED ISSUE #38: Curl(56) error on 9933 when generating session keys
When generating session keys make sure _--unsafe-rpc-external_ is in the docker command, otherwise it will throw a "curl: (56) Recv failure: Connection reset by peer" when operating inside the docker.

Super thanks to @weichweich for the patience in running this one down with me!

weichweich:
## fixes KILTprotocol/docs/issues/38

The container needs to expose it externally so that the port can get mapped to the host port 

## Checklist:


ntn-x2 (approved):
LGTM

PR #88, opened 2022-05-13
feat(concept): distributed trust

weichweich:
## fixes KILTProtocol/ticket#1970

I added the term "attester collectives" to describe the general concept of attesters grouping together and organizing themselves. VCOs and Delegation Hierarchies are examples for that.

todo: 

* [ ] dark theme for SVGs
* [ ] Delegation Hierarchies are inaccurate (things are not mentioned)
    * One can be part of the same hierarchy multiple times
    * the hierarchy is only valid for a single ctype

## Checklist:


docs/concepts/distributed_trust.md
```diff
@@ -0,0 +1,85 @@
...
+
+Attesters can be individuals that attest to the validity of claims made by claimers.
+But in many cases multiple attesters group together to build up a trust in a brand.
+Verifier don't need to trust individual attesters but trust the brand, which in return ensures that all attesters working for this brand are credible.
+
+Such a brand can be organized in many different ways.
+Today the most common way to organize and build a brand is to found a company.
+In the future their are going to be more decentralized solutions like Decentralized Autonomous Organizations (DAOs) or more specifically Virtual Credential Organizations (VCOs).
+
+If such a collective has given the attester the right to issue attestations, the attester can decide to create an attestation in the name of the collective.
```
ntn-x2: Seems a very long description, and I think it's best if it's not the very first thing a reader sees. We don't know yet how attestation works, but we are already talking about when someone leaves the collective.

ntn-x2 (commented):
I like the idea of the collectives, albeit I think this is a very engineer way of describing the concept. Lots of explanations and details which I fear might affect the capability of the reader to follow. Maybe the analogy of a company or even of the traditional CA infrastructure (within the limits) would help understanding.

docs/concepts/distributed_trust.md
weichweich: I think you are right. I wanted to explain a higher level concept so that the reader gets a to know the capabilities of the blockchain and later explain how the Delegation Hierarchies and VCOs implement the specifics.

But it's probably better to keep it even shorter. I would propose:

In the introduction we motivate why attesters should group up and organize themselves. But we don't explain how. After that we have more detailed sections for Delegation Hierarchies and VCOs. There we can go into more details.

[1932479](https://github.com/KILTprotocol/docs/pull/88/commits/19324793b0a4f5d0e9454f72c687d5d8ae82225e)

docs/concepts/distributed_trust.md
```diff
@@ -8,53 +8,29 @@ This feature is currently a preview on Peregrine and is therefore not available
...
+The KILT protocol provides mechanisms to form such brands on the blockchain.
 
-All of these questions are solved by the specific type of collective.
-At the moment there is only a single type of collective available, the Delegation Hierarchy.
-A future attester collective type is the VCO which is more powerful than a simple Delegation Hierarchy and provides more flexible tools.
+There are two ways for attesters to group up and build an organization.
+The first one are Delegation Hierarchies, which provide a very basic hierarchical organisation.
+Delegation Hierarchies are available on our test network Peregrine.
+The second option are Virtual Credential Organizations which are not yet implemented and currently.
+Virtual Credential Organizations will be more flexible than Delegation Hierarchies and accommodate a wider range of use cases.
```
ntn-x2: I would not call it wider, but different. For instance VCOs are unable to represent hierarchies at all, which is something that some use cases might require. I agree on the flexibility thought.

docs/concepts/distributed_trust.md
```diff
@@ -8,53 +8,29 @@ This feature is currently a preview on Peregrine and is therefore not available
...
 
 ## Delegation Hierarchy
 
-Delegation Hierarchies are one type of attester collectives.
-This type of collective organizes it's member strictly hierarchical.
-Every one can create a new hierarchy and imminently become the only member of the collective.
+Delegation Hierarchies organizes it's members strictly hierarchical.
```
ntn-x2: Doesn't this concept, phrased like this, conflict with the idea that a given user can appear twice within the same hierarchy with different roles? I am just suggesting to either remove this sentence or make it a bit more nuanced.

ntn-x2 (commented):
Looks already cleaner and clearer. I think there could be a couple small more changes to avoid confusion 👍

docs/concepts/distributed_trust.md
weichweich: [8074d7f](https://github.com/KILTprotocol/docs/pull/88/commits/8074d7fb0d6a3a6892fefb13609766d6929fa3bd)

docs/concepts/distributed_trust.md
weichweich: Yeah that's right. Didn't think about that. 
[edcacac](https://github.com/KILTprotocol/docs/pull/88/commits/edcacacadbb76a0dd59a39948b7ed3a4ed619b0f)

docs/concepts/distributed_trust.md
```diff
@@ -0,0 +1,67 @@
...
+
+## Delegation Hierarchy
+
+Delegation Hierarchies organize their members in a traditional hierarchical structure.
+Every one can create a new hierarchy and imminently become the only member of the organization.
+Not only is the creator the only member but also the root of the hierarchy.
+They have full control over the Delegation Hierarchy.
+
+When the root adds new attesters to the hierarchy, these new members become direct descendants of the root.
+The root is the parent of the new members while the members are the children of the root.
```
ntn-x2: Seems a bit of a repetition, I would choose only one of them.

docs/concepts/distributed_trust.md
```diff
@@ -0,0 +1,67 @@
...
+  }}
+/>
+
+</center>
+
+The above depicted Delegation Hierarchy contains five attesters.
+Attester 1 is the root of the Delegation Hierarchy.
+As the root Attester 1 added two more attesters.
+Attester 2 was given the right to delegate and to issue attestations for the collective.
+Attester 3 can only add more attesters to the Delegation Hierarchy, but he cannot create, revoke or remove Attestations.
```
ntn-x2, line 51: I think it would be nice to add an example on when this would be desirable. Right now I can't come up with any use cases, but maybe you have some in mind?

docs/concepts/distributed_trust.md
```diff
@@ -0,0 +1,67 @@
...
+Attesters 4 and 5 where added by Attester 3 and can only create, revoke and remove attestations, but they cannot add more attesters to the Delegation Hierarchy.
```
ntn-x2, line 54: I would add an example also for this case, which will probably be more common that the one above. For instance you might want someone to attest on your behalf, if you are on holiday, without the possibility of extending the delegation hierarchy before you return.

docs/concepts/distributed_trust.md
```diff
@@ -0,0 +1,67 @@
...
+
+Delegation Hierarchies also limit who can revoke or remove attestations.
+Only parents of an attester can revoke or remove delegations from their children.
+E.g. Attester 2 cannot revoke/remove attestations issued by Attester 1, 3, 4 and 5, while Attester 1 can revoke and remove attestations issued by any attester since Attester 1 is the root node.
```
ntn-x2:
```suggestion
E.g. Attester 2 cannot revoke/remove attestations issued by Attester 1, 3, 4 and 5, while Attester 1 can revoke and remove attestations issued by any attester since Attester 1 is the root node, hence is the ancestor of every other node.
```

ntn-x2: Or maybe to make it easier:

```suggestion
E.g. Attester 2 cannot revoke/remove attestations issued by Attester 1, 3, 4 and 5, while Attester 1 can revoke and remove attestations issued by any attester since Attester 1 is, directly or indirectly, the parent of every other node.
```

ntn-x2 (commented):
Looks quite nice now! I would still like to see a concrete example at least for the delegation hierarchy, which would include cases of delegating someone with only delegation powers or only attestation powers. This is because not everyone will be familiar with tree and related concepts, so a clear example would be easier to grasp. I guess a company structure would make for the perfect example. What do you think?

docs/concepts/distributed_trust.md
weichweich: The classical use case would be the "manager". They add people to their team but they don't work themselves. 😉

docs/concepts/distributed_trust.md
weichweich: Yes or in a company, you want that the employees who do the attestation work are not able to add people to the company. Only HR or the manager should be able to "hire" new employees.

docs/concepts/distributed_trust.md
weichweich: [3991e27](https://github.com/KILTprotocol/docs/pull/88/commits/3991e272def6b202c0c1a118106e00a7fdbbb6b1)
I want to explain parent & children here. That's why I mention both.

docs/concepts/distributed_trust.md
weichweich: [d5af837](https://github.com/KILTprotocol/docs/pull/88/commits/d5af8378b4d9d20a68b2299fdef41fc3230bde4f)

docs/concepts/distributed_trust.md
weichweich: [d5af837](https://github.com/KILTprotocol/docs/pull/88/commits/d5af8378b4d9d20a68b2299fdef41fc3230bde4f)

ntn-x2 (approved):
LGMT, but I would like a second one to give it a review as well.

PR #95, opened 2022-05-20
docs: remove 'Apps' section and add 'Specifications' section

ntn-x2:
Fixes https://github.com/KILTprotocol/ticket/issues/2018 and fixes https://github.com/KILTprotocol/ticket/issues/2012.

I also renamed `workshop` to `3-workshop` and then added `4-specifications`. The whole way of declaring structure would probably have to be changed before publishing.


docs/develop/4-specifications/01_index.md
```diff
@@ -0,0 +1,20 @@
+---
+id: index
+title: Index
```
tjwelde: I would call it like the menu entry "Technical Specification"

tjwelde (commented):
Not sure about having sidebars just for the sake of it. I would rather not have it, when it only has one item

docs/develop/4-specifications/01_index.md
ntn-x2: Yeah this is supposed to be the entry page for all specifications, but for now it only contains links in the same page. Anyway, I removed the navbar (for now) and changed the page name: https://github.com/KILTprotocol/docs/pull/95/commits/5718b309614acfe35b0bd43027a6b9a1a2e4e20d

tjwelde (approved):
LGTM

PR #98, opened 2022-05-23
docs: correct structure

ntn-x2:
Implement the correct docs structure as outlines in the [draft document](https://hackmd.io/79owvVVhRf2v0xlOVVJ-SA).

Few files have been renamed to always be `01_name` using underscore and always starting from 01.

Kind of a destructive change, but I hope we do it once and get done with it.


tjwelde:
DID you check, that starting with 00 works? In the official docs they always start with 01 and I think using 00 will not put the item first

ntn-x2:
I changed to always start with `01`. There's no `00` left.

Updated the PR description to match that 😅

tjwelde:
Also using metadata for the autogenerated sidebars is the recommended way: https://docusaurus.io/docs/sidebar/autogenerated#doc-item-metadata
So I was thinking, if we should rather move in that direction...

tjwelde (approved):
LGTM though, definitely better than before

wischli (approved):
The folder naming looks weird now (e.g. `04_Issue and verify a KILT credential`) as it can't follow the underscore convention. However, I have no suggestion on how to improve it without breaking the compiler or frontend display and this PR definitely adds value.

IMO, we should look into autogenerated sidebars but for now: LGTM!

ntn-x2:
Agree. We will look into metadata for better experience. Merging this one now 👍

PR #103, opened 2022-05-27
chore: major refactoring of test cases

ntn-x2:
Fixes https://github.com/KILTprotocol/ticket/issues/2024 and fixes https://github.com/KILTprotocol/ticket/issues/2028.

Important upgrade is that it brings testing time for core features from ~8 minutes down to around 6. It could be improved further by parallelising tests also within each flow, but I think this is good enough for now.

It also fixes all the links, and changes the quickstart from using several small scripts, to simple reference lines in the only snippet, which is then also shown in full at the end of the quickstart.


tjwelde:
Not sure about the `SnippetBlock`, since when the original file changes, we would not be informed about it

wischli (approved):
LGTM!

tjwelde (approved):
LGTM

PR #102, opened 2022-05-27
feat: add tips section

wischli:
* Adds section about tips which are part of the next Spiritnet runtime upgrade (targeted mid June).
* Part of [KILT Ticket #2026](https://github.com/KILTprotocol/ticket/issues/2026)

PR #96, opened 2022-05-23
feat: add treasury proposal page from the homepage to the docs

trusch:
## fixes KILTProtocol/ticket#2014

This adds the section "Treasury Proposal" to the participate section.
The content comes from the page on the homepage https://www.kilt.io/treasury/create-a-proposal/

## How to test:
Please provide a brief step-by-step instruction.
If necessary provide information about dependencies (specific configuration, branches, database dumps, etc.)

- Step 1
- Step 2
- etc.

## Checklist:

- [x] I have verified that the code works
- [x] I have verified that the code is easy to understand
  - [ ] If not, I have left a well-balanced amount of inline comments
- [ ] I have [left the code in a better state](https://deviq.com/principles/boy-scout-rule)
- [ ] I have documented the changes (where applicable)

PR #53, opened 2022-02-22
feat: contribution guide

Dudleyneedham:
## fixes https://github.com/KILTprotocol/ticket/issues/1847

Adds the contribution

## How to test:

yarn start

## Checklist:

- [ ] I have verified that the code works
- [ ] I have verified that the code is easy to understand
  - [ ] If not, I have left a well-balanced amount of inline comments
- [ ] I have [left the code in a better state](https://deviq.com/principles/boy-scout-rule)
- [ ] I have documented the changes (where applicable)

PR #107, opened 2022-05-31
feat: add propose tip section

wischli:
* Additionally to #102, this PR adds a section on how to propose a tip and the lifecycle
* Partially dependent on #53 and #96, as references should be linked
* Regarding the example: After I noticed that the example on the [Polkadot Wiki](https://wiki.polkadot.network/docs/learn-treasury#example) was exactly, what I was trying to produce, I felt better referencing to it. If you think, we should add one of our own, I am fine with doing so.
* Note: Final images of the extrinsic and navigation will be added once tips are live on Spiritnet


docs/participate/03_Contribute/02_propose-tip.md
```diff
@@ -0,0 +1,98 @@
...
+Of course, success of such can only be expected if it is based on a meaningful contribution.
+<!-- TODO: Add link to Contribution guide once #53 is merged -->
+The variety of potential contributions is vast, please see our Contribution guide for more details.
+There, you will also find a high level description of tips and the differences to Treasury proposals.
+
+In the following, we will lead you through the necessary steps from requesting a tip to finally receiving it.
+
+## Lifecycle of a tipping process
+
+Since anyone propose as tip, you can certainly do that for someone else, the **<span style={{background: '#7cd27c', color: 'black'}}>Beneficiary</span>**, as well.
```
wischli: I thought about moving these styles into a css file but since they are used exactly once, I did not go with it.

trusch (approved):

PR #108, opened 2022-05-31
feat: test print json examples

tjwelde:
Fixes https://github.com/KILTprotocol/ticket/issues/2044.

This PR adds a preliminary step when building the docs which runs a simple script, creates all the examples that are needed throughout the docs, and saves their JSON representation in some files at a well known location.

The benefits from this solution are multiple:

- Snippets are dynamically generated, hence always up to date
- If anything changes in the SDK, the script would break and we would notice it
- Beside few things like the nonce in the encrypted message example, since the generated files are also part of the Git history, it's possible to see if there has been any breaking change in the structure of these snippets

Another difference from the solution @tjwelde envisioned, is that everything is rendered server-side when the project is compiled, meaning that no runtime, client-side errors are expected to be generated if the build process completes successfully.

## How to test

Just run `yarn build` from the root directory: it will now call `yarn generate-examples`, which `cd`s into the `scripts` directory and run the generation script, which typically takes less than 10 seconds.


tjwelde:
Weirdly I couldn't get it to work with SSR, so the example is rendered in the browser, instead of pre-rendered on the server-side.

tjwelde:
Naming is also not good yet and stuff might be needed to move around, but it shows how it would work

docs/concepts/04_credentials/02_ctypes.md
```diff
@@ -2,6 +2,11 @@
 id: ctypes
 title: CTypes
 ---
+import CodeBlock from '@theme/CodeBlock';
+
+<!-- Taken from https://github.com/webpack-contrib/raw-loader/issues/91#issuecomment-648830498 -->
+import ctypeSchema from '@site/scripts/out/ctype-schema.json.raw!=!raw-loader!@site/scripts/out/ctype-schema.json';
```
tjwelde, line 8: what does the `from` do here? Would it not have been enough to just `import ctypeSchema from '!raw-loader...'`

docs/develop/01_sdk/04_integrate/01_nodejs.md
```diff
@@ -6,15 +6,10 @@ title: NodeJS
 import CodeBlock from '@theme/CodeBlock';
 import QueryAccountName from '!!raw-loader!@site/code_examples/core_features/linking/03_account_web3name_query.ts';
 
-
 NodeJS is natively supported and doesn't require any additional setup.
 
 Have a look at these example `package.json` and `index.js` files for reference:
 
-<!--
```
tjwelde, line 14: This we don't want to dynamically generate anymore?

tjwelde (commented):
Some questions

docs/concepts/04_credentials/02_ctypes.md
ntn-x2: Apparently there is a bug with loading raw JSON files, as mentioned in the link I added in the comment. It does not work if the file is loaded "regularly".

docs/develop/01_sdk/04_integrate/01_nodejs.md
ntn-x2: That is just a `package.json`. I updated the CI job to change SDK versions in this folder as well, so that would be enough I feel. I don't see much value in dynamically generating a config file, since it would also then require to spin up a test project and see if it can be created correctly. I felt it is out of scope for this PR, if we want to tackle that at all.

tjwelde:
LGTM ;)

code_examples/core_features/web3names/03_query_name_credentials.ts
```diff
@@ -78,7 +78,7 @@ export async function queryPublishedCredentials(
   // The case where the result is not a JSON should be properly handled in production settings.
   const credentialCollection: CredentialEntry[] = await fetch(
     firstCredentialCollectionEndpointUrl
-  ).then((response) => response.json())
+  ).then((response) => response.json().then((r) => r as CredentialEntry[]))
```
rflechtner: The extra then just for the sake of the type system is unnecessary I think. If you really want/must add extra type annotation, use
```suggestion
  ).then((response) => (response.json() as Promise<CredentialEntry[]>))
```

.github/workflows/sdk-update-test.yml
```diff
@@ -91,14 +91,25 @@ jobs:
           include: "code_examples/**"
           exclude: "**/*.lock"
           regex: true
-      - name: Regenerate yarn lock
+      - uses: jacobtomlinson/gha-find-replace@v2
```
rflechtner, line 115: pls don't remove the vitejs locking when fixing merge conflicts

docs/concepts/06_messaging.md
```diff
@@ -26,15 +30,9 @@ As long as both parties keep their secret keys well protected, the combination o
...
-```ts
-{
-  receiverKeyUri: 'did:kilt:4qMx…66m4#0x4af3…cdcb' // indicates the key to be used when decrypting
-  senderKeyUri: 'did:kilt:4qtP…m507#0x65ac…a282' // indicates the key used to encrypt
-  ciphertext: '0xdeadbeef…' // message contents in unreadable, encrypted form
-  nonce: '0x1234…' // random data required for decryption
-}
-```
+<CodeBlock className="language-json">
+  {encryptedMessage}
```
rflechtner, line 34: The cipher text seems a bit much, but I guess that's just the price we pay for proper json examples.

package.json
```diff
@@ -41,4 +42,4 @@
       "last 1 safari version"
     ]
   }
-}
+}
```
rflechtner:
```suggestion
}

```

scripts/package.json
```diff
@@ -0,0 +1,29 @@
...
+  "license": "MIT",
+  "scripts": {
+    "lint": "eslint . --ext .ts --format=codeframe",
+    "lint:fix": "yarn lint --fix",
+    "style": "prettier --check --config .prettierrc '**/*.ts'",
+    "style:fix": "yarn style --write",
+    "generate": "ts-node src/main.ts"
+  },
+  "dependencies": {
+    "@kiltprotocol/sdk-js": "0.28.0"
```
rflechtner:
```suggestion
    "@kiltprotocol/sdk-js": "0.28.1"
```

rflechtner (approved):
Just some nitpicks, love it!

docs/concepts/06_messaging.md
ntn-x2: Yeah indeed. I could change the object properties after it has been generated, but I think this is more indicative.

.github/workflows/sdk-update-test.yml
ntn-x2: I will go ahead and merge this, but can you maybe double check when you're back that I've done it correctly?

rflechtner: looks good!

PR #109, opened 2022-06-01
Adds logic to update docs only when official sdk is released

ggera:
## NO TICKET
Do not open PR when sdk is not official release

## Checklist:


ntn-x2 (approved):
👌

PR #125, opened 2022-07-19
fix(deps/ci): add node-fetch@^2 as direct dependency, fix frozen install

rflechtner:
## quick fix 

Node-fetch was used in a test file but not added as a direct dependency, resulting in a failure to run tests when it was bumped to v3 by the dependency that required it ([here](https://github.com/KILTprotocol/docs/runs/7391749275?check_suite_focus=true)). Added `node-fetch@^2.6.7` as a direct dev-dependency to resolve.

Additionally, refreshing the lock files revealed that they were out of sync with their respective `package.json`. This went unnoticed because the workflows used the command `yarn install --frozen`, but the `--frozen` option does not exist. Fixed that too (it's `--frozen-lockfile`). 

## How to test:

I can revert the lock file refresh to prove my point I guess (see https://github.com/KILTprotocol/docs/actions/runs/2698400017).
Also checked that the issue with the node-fetch import is now resolved.

## Checklist:

### DIFF
--- a/.github/workflows/deploy-docs.yml
+++ b/.github/workflows/deploy-docs.yml
@@ -26,5 +26,5 @@ jobs:
         run: |
           git config --global user.email "actions@github.com"
           git config --global user.name "gh-actions"
-          yarn install --frozen-lockfile
+          yarn install --frozen-lockfile-lockfile
           yarn run deploy
--- a/.github/workflows/deploy-docs.yml
+++ b/.github/workflows/deploy-docs.yml
@@ -26,5 +26,5 @@ jobs:
         run: |
           git config --global user.email "actions@github.com"
           git config --global user.name "gh-actions"
-          yarn install --frozen-lockfile-lockfile
+          yarn install --frozen-lockfile
           yarn run deploy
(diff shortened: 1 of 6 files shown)


.github/workflows/deploy-docs.yml
```diff
@@ -26,5 +26,5 @@ jobs:
         run: |
           git config --global user.email "actions@github.com"
           git config --global user.name "gh-actions"
-          yarn install --frozen-lockfile
+          yarn install --frozen-lockfile-lockfile
```
rflechtner:
```suggestion
          yarn install --frozen-lockfile
```

code_examples/core_features/package.json
```diff
@@ -15,6 +15,8 @@
     "dotenv": "^16.0.0"
   },
   "devDependencies": {
+    "node-fetch": "^2.6.7",
+    "@types/node-fetch": "2.6.2",
```
ntn-x2: `@types/node-fetch` seems to be pinned to an exact version. How do we make sure it does not break if `node-fetch` changes something? Is this something we should/could even take care of?

ntn-x2 (approved):
LGTM, good catch! I had just a small remark/question

code_examples/core_features/package.json
rflechtner: oh, yeah that was not what I wanted, good catch

PR #126, opened 2022-07-20
fix: prettier style script

rflechtner:
## quick fix

The `yarn style` command was broken; the glob pattern needs to be quoted to work in the package script. Therefore the results of running the command directly from the console and via the package script diverged. Now style errors are detected and fixed by the style script again. 

## How to test:

See workflow run for commit 329444a49544a0e028cd6210ade3a0d4fe1ad9ba vs commit https://github.com/KILTprotocol/docs/commit/7d90277a7e43867c5e0d7377c99deef4e86bd159

## Checklist:


ntn-x2 (approved):
Interesting, I never had that error before. Anyway, good catch!

PR #128, opened 2022-07-20
fix: workshop section order made no sense

rflechtner:
## quick fix

We have a premature conclusion in the workshop that preceded the verification section. I presume this was meant to be the other way around and thus fixed it.

## How to test:

Compare live workshop to the deployment on this PR, I guess.

## Checklist:


ntn-x2 (approved):
👌👌👌

PR #131, opened 2022-07-27
fix(ci): sdk release triggered workflow

rflechtner:
## fixes KILTprotocol/ticket#2119
The sdk release test pipeline is broken bc someone deleted the develop branch without changing the PR target in the wf.
Also we don't refresh the lock file of the vitejs code examples.

## How to test:
I honestly don't know how to test this, but it's definitely less broken than before.

## Checklist:


ntn-x2 (approved):
All good!

PR #146, opened 2022-09-09
feat: update docs to latest SDK v0.29

ntn-x2:
PR created to get an Amplify deployment of what will go into `master` once all changes are in.

Fixes https://github.com/KILTprotocol/ticket/issues/2189.

## STILL TO DO

- [x] Add support for deprecating the old docs and point to the new ones
- [x] Upgrade to the official SDK version once released


docs/develop/07_dApp/02_session.md
```diff
@@ -25,25 +30,9 @@ The dapp introduces itself to the extension with its name, encryption key URI, a
...
-
-const session = await window.kilt.sporran.startSession(
-  dAppName,
-  dAppEncryptionKeyUri,
-  challenge
-)
-```
+<TsJsSnippet>
+  {DappIntroduction}
+</TsJsSnippet>
```
weichweich, line 35: The imports are missing from the code example. Would be helpful to include them. :innocent:

ntn-x2: The thing is that before the function body there are few declared variables that make the whole TS file compile. If we show the whole file, including the imports, these would be shown and it is not good. We don't have component that can reliably exclude some lines from the output, at the moment. We can either show the whole file, or a function body given its name and its start and end delimiter.

rflechtner (approved):
👍

PR #147, opened 2022-09-09
feat: translate TS snippets to JS

ntn-x2:
Fixes https://github.com/KILTprotocol/ticket/issues/2162.

It adds one more step to the build process, which translates all TS snippets to JS, which can then be imported throughout the docs.
It also introduces two custom components, `TsJsSnippet` and `TsJsBlock` which can take a TS and a JS snippet and build a `<Tabs>` component with the right details (component is very basic, it could be improved to add, for instance, props specific to each TS or JS state).

Only downside is that snippets have to be imported two times in the docs, and sometimes even the text has to be duplicated, if put inside a `<Tabs>` variant. The docs could be improved so to minimize language-specific stuff (like file names, for instance), but that is out of scope of this PR.


tjwelde:
Would be good, if we can gitignore the js files

docs/develop/01_sdk/01_quickstart.md
```diff
@@ -4,8 +4,7 @@ title: Quickstart
 ---
 import CodeBlock from '@theme/CodeBlock';
 import SnippetBlock from '@site/src/components/SnippetBlock';
-import Tabs from '@theme/Tabs';
```
tjwelde, line 7: `<Tabs>` and `<TabItem>` are used in this file. Do we not have to import those anymore?

docs/develop/03_workshop/02_setup.md
```diff
@@ -3,8 +3,6 @@ id: setup
 title: 🎒 Setup
 ---
 
-import Tabs from '@theme/Tabs';
```
tjwelde, line 6: Same as above. We use it still in this file, but is not imported anymore? curious...

src/components/TsJsBlock/index.js
```diff
@@ -0,0 +1,48 @@
...
+
+const TsJsBlock = ({ children, fileName, ...props }) => {
+  const tsSnippet = children
+  // 1. Transpile TS to JS
+  const { code: jsSnippet } = transform(tsSnippet, {
+    plugins: ['transform-typescript'],
+    retainLines: true,
+  })
+  // 2. Prettify the resulting JS
+  // FIXME: Maybe we can pass this as a config to the prop?
```
tjwelde: You could use `useContext` for sending the config down from the root of the app, if you want

src/components/TsJsBlock/index.js
```diff
@@ -0,0 +1,48 @@
...
+  const prettyJsSnippet =
+    format(
```
tjwelde, line 23: Love the combination of retainLines and prettier

tjwelde (approved):
The result looks _very_ good now! love it!

docs/develop/01_sdk/01_quickstart.md
ntn-x2: I guess once the component is imported once, it is always there. There was no warning whatsoever either. But I fixed that, thanks! https://github.com/KILTprotocol/docs/pull/147/commits/7aef98cc92d0acdc1243d70fc8db956147bdf67c

docs/develop/03_workshop/02_setup.md
ntn-x2: https://github.com/KILTprotocol/docs/pull/147/commits/7aef98cc92d0acdc1243d70fc8db956147bdf67c

src/components/TsJsBlock/index.js
ntn-x2: Did that, thank you! https://github.com/KILTprotocol/docs/pull/147/commits/5863573ff78e35cdcf483987c4b97837d7996d02

Did not find a way to open the `.prettierrc` file directly, so there is a bit of duplication. Do you think it's fine?

tjwelde: I think the duplication is fine. I would suggest putting the prettierConfig into the customFields as a separate property though, in  case want to add more custom fields in the future
```
  customFields: {
    prettierConfig: {
      trailingComma: "es5",
      semi: false,
      singleQuote: true,
      printWidth: 80
    }
  }
```

docs/develop/01_sdk/01_quickstart.md
tjwelde: ah ok. yeah not sure how the imports work on the mdx files. But doesn't hurt to import them i guess

PR #152, opened 2022-09-15
feat: update snippet to use the format for published credentials

ntn-x2:
Fixes https://github.com/KILTprotocol/ticket/issues/2036.

Uses the standard as defined in the spec https://github.com/KILTprotocol/spec-KiltPublishedCredentialCollectionV1.

## NOTE

This PR is against `master`, as the changes must be integrated live already, since the `john_doe` DID already has the new additional endpoint. Changes will then be merged on `develop` again as part of https://github.com/KILTprotocol/docs/pull/146.

### DIFF
--- a/code_examples/core_features/getting_started/06_fetch_endpoint_data.ts
+++ b/code_examples/core_features/getting_started/06_fetch_endpoint_data.ts
@@ -15,8 +15,8 @@ export async function main(
     metadata?: any
   }
   const {
-    data: { credential }
-  } = await axios.get<PublishedCredential>(endpoints[1].urls[0])
+    data: [{ credential }]
+  } = await axios.get<PublishedCredential[]>(endpoints[1].urls[0])
 
   return credential
 }
--- a/code_examples/core_features/getting_started/06_fetch_endpoint_data.ts
+++ b/code_examples/core_features/getting_started/06_fetch_endpoint_data.ts
@@ -10,14 +10,13 @@ export async function main(
 ): Promise<Kilt.IRequestForAttestation> {
   // Define the type of a published credential
   type PublishedCredential = {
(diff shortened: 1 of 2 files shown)


code_examples/core_features/getting_started/06_fetch_endpoint_data.ts
```diff
@@ -1,14 +1,22 @@
...
-    .then((response) => response.data)
+  // Define the type of a published credential
+  type PublishedCredential = {
+    credential: Kilt.IRequestForAttestation
+    // Not relevant for this case, but production applications should parse this field as well
+    metadata?: any
+  }
+  const {
+    data: [{ credential }]
+  } = await axios.get<PublishedCredential[]>(endpoints[1].urls[0])
```
wischli, line 19: Are you intentionally querying the 2nd endpoint or did you mean to query the first one?

```suggestion
  } = await axios.get<PublishedCredential[]>(endpoints[0].urls[0])
```

wischli (commented):
One minor question. Apart from that LGTM for an outsider without recent SDK knowledge.

code_examples/core_features/getting_started/06_fetch_endpoint_data.ts
ntn-x2: Yes, Timo suggested to maintain the old endpoint, so that w3n.id would not break, and add a second endpoint with the new format.

ntn-x2: The idea would be to remove the old one once at least w3n.id and DIDSign have bene updated. Which we can push for with the new SDK release.

tjwelde (approved):
LGTM

PR #153, opened 2022-09-15
feat: embrace async/await more

arty-name:
Removes last remaining `.then()` calls


ntn-x2 (approved):
Apart from the failing tests of coding convention (which you can fix with `yarn fix` in each of the three code example folders) LGTM!

PR #154, opened 2022-09-15
feat: use new subscription promise settings

ntn-x2:
Much needed addition!

Tests for the core features will fail since Spiritnet published credential has changed to the new format, and changes from https://github.com/KILTprotocol/docs/pull/152 must be pulled in first.


arty-name (approved):
Very nice!

PR #157, opened 2022-09-19
fix: vite.js build

weichweich:
## fixes KILTprotocol/ticket/issues/2201

fix the polyfills for production builds

## Checklist:


trusch (approved):
lgtm

ntn-x2 (commented):
Can we add a package.json script that runs both commands? And then add it to the build pipeline for the vitejs part? I think it would be very nice.

PR #160, opened 2022-09-26
chore: update SDK to v0.29.0-rc.2

ntn-x2:
Mostly code snippets have been updated. A different PR will be opened for all the doc text.


code_examples/core_features/did/01_light_did_simple.ts
```diff
@@ -1,21 +1,13 @@
...
-): Kilt.DidDocument {
-  // Ask the keyring to generate a new keypair to use for authentication with the generated seed.
-  const authKey = keyring.addFromSeed(
-    authenticationSeed
-  ) as Kilt.NewLightDidVerificationKey
-
+export function createSimpleLightDid({
+  authentication
+}: {
+  authentication: Kilt.KiltKeyringPair
```
arty-name: Would it work like this? We could skip the type casting below in this case.
```suggestion
  authentication: Kilt.NewLightDidVerificationKey
```

arty-name (commented):
I still have lots to review, but this idea might reduce type casting across the code

code_examples/core_features/did/01_light_did_simple.ts
ntn-x2: But if the function returns a DID key, we can't use it for a sign callback, right? The key does not have a `.sign()` function.

arty-name: I meant this part:
```ts
  const lightDID = Kilt.Did.createLightDidDocument({
    authentication: [authentication as Kilt.NewLightDidVerificationKey]
  })

Comment (ntn-x2) [reply code_examples/core_features/did/01_light_did_simple.ts]: Do you think this is a good fix? https://github.com/KILTprotocol/docs/pull/160/commits/e338aebd35982e00a49aad77f30e0446e9faeb78

I wanted to keep parameters for light and full DID creation the same, but that means that for a full DID creation also the sign callback must be given now, as the key cannot be used to sign the payload directly. What do you think?

File: code_examples/core_features/run_core_features.ts [hunk 2]
```diff
@@ -123,7 +156,11 @@ async function main(): Promise<void> {
     await main()
     process.exit(0)
   } catch (e) {
-    console.log('Error in the core features test', e)
+    if (e instanceof Error) {
+      console.error(e.message)
```
arty-name: I think this way it will not print the helpful stacktrace

code_examples/workshop/attester/generateAccount.ts
```diff
@@ -1,47 +1,34 @@
 import { config as envConfig } from 'dotenv'
 
-import { cryptoWaitReady, mnemonicGenerate } from '@polkadot/util-crypto'
-import { Keyring } from '@polkadot/api'
+import { mnemonicGenerate, mnemonicToMiniSecret } from '@polkadot/util-crypto'
 
 import * as Kilt from '@kiltprotocol/sdk-js'
 
-export function generateAccount(keyring: Keyring): {
+export function generateAccount(mnemonic?: string): {
```
arty-name: Would this be more straightforward?
```suggestion
export function generateAccount(mnemonic = mnemonicGenerate()): {
```

arty-name (approved):
I wonder if we even still need some of the simpler `generateKeypairs()` functions

code_examples/core_features/run_core_features.ts
ntn-x2: Any suggestions? Should we also print the stack trace? I found that the CI needs the exception to be caught and processed somehow. I tried without and it was complaining that it should be caught.

ntn-x2: Any suggestions? Should we also print the stack trace? I found that the CI needs the exception to be caught and processed somehow. I tried without and it was complaining that it should be caught.

arty-name: A simple `console.error(e)` would work better, I think

arty-name: In other places as well

PR #172, opened 2022-10-26
fix: link checking

rflechtner:
## fixes KILTProtocol/ticket#2240

Checks all the links in markdown files. The problem resides in glob extension working differently in zsh as compared to bash / sh. While zsh recursively matches all files, sh used by yarn to run package scripts considers `/**/` to indicate exactly one level of nesting. Fixed this using `find`. Also explored a solution using the npm package `glob-exec`.

This PR additionally fixes a few broken links we've overlooked so far.

## How to test:

Break a link!

## Checklist:


ntn-x2 (approved):
Love it! 😍😍😍

PR #173, opened 2022-10-26
feat: make sure example generation is deterministic

rflechtner:
Hack nonce generation to make example generation deterministic for easier detection of changes.


ntn-x2 (approved):
Good point! Perhaps we can update the SDK to take an optional nonce as well, in the future.

rflechtner:
> Good point! Perhaps we can update the SDK to take an optional nonce as well, in the future.

Mhh I’m afraid that would invite people to do stupid shit and break encryption. I’m fine with hacking it for cases like this

PR #176, opened 2022-11-07
Update generateCtype.ts

niklasp:
to display message on successful creation

## Checklist:


ntn-x2:
Hi! Thanks for the PR. Do you think this information is necessary? I think printing a CType can clog the output quite a lot. Why do you think you need to print it out?

niklasp:
When i was running the example i was missing visual feedback if the operation was successful.

ChrisChinchilla (approved):
For this workshop, I think the output is useful, and will fix merge issues.

PR #179, opened 2022-11-07
fix verify example and add fileNames to code example

niklasp:
## fixes not working verify.ts code example
add a missing `Kilt.connect()` and add file names to code examples in verify view.

## How to test:
Before this patch run `yarn ts-node verify.ts` and get the error 
```
BlockchainApiMissingError: The blockchain API is not set. Did you forget to call Kilt.connect(…) or Kilt.init(…)?
```

after this patch it works and was tested

## Checklist:


ntn-x2:
Thank you! I moved the missing `connect()` inside the try-catch block to make it consistent with the other occurrences.

PR #181, opened 2022-11-16
feat: SignCallback

weichweich:


docs/develop/01_sdk/02_cookbook/06_signCallback.md
```diff
@@ -0,0 +1,78 @@
...
+Storing the private key on a separate device or inside a sandboxed application can increase security.
+But to enable these security options, we need a generic interface to talk to the signer.
+This is what the `SignCallback` does.
+
+The `SignCallback` defines an interface between the SDK and an arbitrary signing strategy.
+May it be a ledger, an air gaped phone or your browser extension.
+The interface is generic enough to support implementations for all these security measures.
+There are three different flavours of the `SignCallback`.
+
+## The family of SignCallbacks
```
Dudleyneedham:
```suggestion
## SignCallbacks Family
```

code_examples/core_features/claiming/index.ts
```diff
@@ -17,14 +17,13 @@ export async function runAll(
   submitterAccount: Kilt.KiltKeyringPair
 ): Promise<void> {
   console.log('Running claiming flow...')
-  let { authentication, encryption, attestation, delegation } =
-    generateDidKeypairs()
+  const lightDidAuthKey = generateDidKeypairs().authentication
```
Dudleyneedham: It would be better to deconstruct it completely here as it was before.

I don't know, if we should reassign to give the new lightDidAuthKey.

Do you think it will confuse people if we just make it the Authentication key?

code_examples/core_features/run_core_features.ts
```diff
@@ -112,42 +113,17 @@ async function main(): Promise<void> {
...
-          nodeAddress,
-          accountLinkingTestAccount,
-          faucetAccount
-        )
-      } catch (e) {
-        console.error('Linking flow failed')
-        throw e
-      }
-    })()
+    runAllClaiming(claimingTestAccount).catch(() =>
```
Dudleyneedham: This looks and feels cleaner

code_examples/core_features/signCallback/getStoreTxSignCallback.ts
```diff
@@ -0,0 +1,13 @@
+import * as Kilt from '@kiltprotocol/sdk-js'
+import { generateDidKeyPair } from '.'
+
+export const getStoreTxSignCallback: Kilt.Did.GetStoreTxSignCallback = async ({
```
Dudleyneedham: Can you not just import the type `GetStoreTxSignCallback`? In other places, we import things individually.

code_examples/core_features/signCallback/getStoreTxSignCallback.ts
```diff
@@ -0,0 +1,13 @@
+import * as Kilt from '@kiltprotocol/sdk-js'
```
Dudleyneedham:
```suggestion
import { GetStoreTxSignCallback } from '@kiltprotocol/sdk-js'
```

code_examples/core_features/signCallback/getStoreTxSignCallback.ts
Dudleyneedham:
```suggestion
export const getStoreTxSignCallback: GetStoreTxSignCallback = async ({
```

code_examples/core_features/signCallback/index.ts
```diff
@@ -0,0 +1,53 @@
+import * as Kilt from '@kiltprotocol/sdk-js'
```
Dudleyneedham, line 1: Why do we have this twice.

code_examples/core_features/signCallback/signCallback.ts
```diff
@@ -0,0 +1,19 @@
...
+import { lookupDidDocument, lookupDidKeyPair } from '.'
+
+export const signCallback: SignCallback = async ({
+  data,
+  keyRelationship,
+  did
+}) => {
+  // look up the DID document for the DID that should sign the data
+  const didDoc = lookupDidDocument(did)
+  const keyUri: DidResourceUri = `${did}${didDoc[keyRelationship][0].id}`
```
Dudleyneedham: Can you change this to be `didUri`? Since it is more explicit. It helps people know it is only the URI and not the whole DID.

code_examples/core_features/signCallback/signExtrinsicCallback.ts
```diff
@@ -0,0 +1,19 @@
+import { DidResourceUri, SignExtrinsicCallback } from '@kiltprotocol/sdk-js'
+import { lookupDidDocument, lookupDidKeyPair } from '.'
+
+export const signExtrinsicCallback: SignExtrinsicCallback = async ({
+  data,
+  keyRelationship,
+  did
+}) => {
+  // Look up the DID keypair for the given key relationship.
+  const didDoc = lookupDidDocument(did)
```
Dudleyneedham: Same again.

code_examples/core_features/utils/generateKeypairs.ts
```diff
@@ -1,8 +1,16 @@
 import { mnemonicGenerate, mnemonicToMiniSecret } from '@polkadot/util-crypto'
 
 import * as Kilt from '@kiltprotocol/sdk-js'
```
Dudleyneedham, line 3: Can you make a single import please?

docs/develop/01_sdk/02_cookbook/06_signCallback.md
```diff
@@ -0,0 +1,78 @@
...
+title: SignCallback
+---
+
+import TsJsBlock from '@site/src/components/TsJsBlock';
+
+import SignCallback from '!!raw-loader!@site/code_examples/core_features/signCallback/signCallback.ts';
+import SignExtrinsicCallback from '!!raw-loader!@site/code_examples/core_features/signCallback/signExtrinsicCallback.ts';
+import GetStoreTxSignCallback from '!!raw-loader!@site/code_examples/core_features/signCallback/getStoreTxSignCallback.ts';
+
+Signing data involves the private key and therefore needs to be secured.
```
Dudleyneedham:
```suggestion
Signing data involves using the private key and therefore needs to be secure.
```

code_examples/core_features/claiming/index.ts
weichweich: This code is not shared publicly as far as I know. It's only used for internal testing.

To be honest, i thought the way we did it before was confusing. We used the same variables for the claimer keys and later for the attester key.

code_examples/core_features/signCallback/getStoreTxSignCallback.ts
weichweich: Good point I very shortly thought about that and then forgot it.

The reason for importing it like that is, that it couldn't get resolved automatically by vscode. I just copied it over from somewhere else. And there it was imported like that.

code_examples/core_features/signCallback/getStoreTxSignCallback.ts
weichweich: `Module '"@kiltprotocol/sdk-js"' has no exported member 'GetStoreTxSignCallback'.`

code_examples/core_features/claiming/index.ts
Dudleyneedham: If they are for different actors, then it is fine. I would just suggest calling it as the claimer.

code_examples/core_features/signCallback/index.ts
weichweich: why twice? It's only there once?

Dudleyneedham: You have import from the SDK `* as KILT from '@kiltprotocol/sdk-js` then you import other things from the  '@kiltprotocol/sdk-js'.

code_examples/core_features/utils/generateKeypairs.ts
weichweich: [0856796](https://github.com/KILTprotocol/docs/pull/181/commits/085679676012d8e6e487cce24c4d518198cef4a2)

code_examples/core_features/signCallback/signExtrinsicCallback.ts
weichweich: [0856796](https://github.com/KILTprotocol/docs/pull/181/commits/085679676012d8e6e487cce24c4d518198cef4a2)

code_examples/core_features/signCallback/signCallback.ts
weichweich: [0856796](https://github.com/KILTprotocol/docs/pull/181/commits/085679676012d8e6e487cce24c4d518198cef4a2)

code_examples/core_features/signCallback/index.ts
weichweich: got it [0856796](https://github.com/KILTprotocol/docs/pull/181/commits/085679676012d8e6e487cce24c4d518198cef4a2)

code_examples/core_features/signCallback/getStoreTxSignCallback.ts
weichweich: [0856796](https://github.com/KILTprotocol/docs/pull/181/commits/085679676012d8e6e487cce24c4d518198cef4a2)

code_examples/core_features/claiming/index.ts
weichweich: [c564291](https://github.com/KILTprotocol/docs/pull/181/commits/c5642919c1c49dab296db484b9360200a3f7b7cb)

docs/develop/01_sdk/02_cookbook/06_signCallback.md
```diff
@@ -0,0 +1,78 @@
...
+---
+
+import TsJsBlock from '@site/src/components/TsJsBlock';
+
+import SignCallback from '!!raw-loader!@site/code_examples/core_features/signCallback/signCallback.ts';
+import SignExtrinsicCallback from '!!raw-loader!@site/code_examples/core_features/signCallback/signExtrinsicCallback.ts';
+import GetStoreTxSignCallback from '!!raw-loader!@site/code_examples/core_features/signCallback/getStoreTxSignCallback.ts';
+
+Signing data involves using the private key and therefore needs to be secure.
+There are many different options how data could be sign.
```
ntn-x2:
```suggestion
There are many different options how data could be signed.
```

docs/develop/01_sdk/02_cookbook/06_signCallback.md
```diff
@@ -0,0 +1,78 @@
...
+You might have the private key stored in memory and are therefore able to simply sign the data.
+This is the easiest option but also comes with higher security risk.
+Storing the private key on a separate device or inside a sandboxed application can increase security.
+But to enable these security options, we need a generic interface to talk to the signer.
+This is what the `SignCallback` does.
+
+The `SignCallback` defines an interface between the SDK and an arbitrary signing strategy.
+May it be a ledger, an air gaped phone or your browser extension.
```
ntn-x2:
```suggestion
May it be a ledger, an air gapped phone or your browser extension.
```

docs/develop/01_sdk/02_cookbook/06_signCallback.md
```diff
@@ -0,0 +1,78 @@
...
+The interface is generic enough to support implementations for all these security measures.
+There are three different flavours of the `SignCallback`.
+
+## SignCallbacks Family
+
+There are three types of signing callbacks.
```
ntn-x2: Repetition of the two lines above.

docs/develop/01_sdk/02_cookbook/06_signCallback.md
```diff
@@ -0,0 +1,78 @@
...
+The `SignCallback` is the most general and can be used in most all cases, except when a DID is registered on chain.
```
ntn-x2:
```suggestion
The `SignCallback` is the most general and can be used in almost all cases, except when signing a full DID creation transaction.
```

docs/develop/01_sdk/02_cookbook/06_signCallback.md
```diff
@@ -0,0 +1,78 @@
...
+The `SignExtrinsicCallback` is a special `SignCallback` which can only be used to sign extrinsics.
+Every `SignCallback` can also be used as a `SignExtrinsicCallback`.
+The `GetStoreTxSignCallback` can only be used to sign the registration of a new DID.
```
ntn-x2: I have never seen the term registration being used, so we should be consistent.
```suggestion
The `GetStoreTxSignCallback` can only be used to sign the creation of a new DID.
```

docs/develop/01_sdk/02_cookbook/06_signCallback.md
```diff
@@ -0,0 +1,78 @@
...
+
+### SignCallback
+
+The plain `SignCallback` signs arbitrary data.
```
ntn-x2, line 34: We have a way to generate JSON files dynamically at build time, so perhaps we could use that here, and then add text below it? So in case anything changes in the future, we can catch the change and update the docs accordingly.

docs/develop/01_sdk/02_cookbook/06_signCallback.md
```diff
@@ -0,0 +1,78 @@
...
+It is called with `SignRequestData` which contains
+
+* the `data` (UInt8Array) that should be signed
+* the `keyRelationship` which specifies which DID key must be used
+* and the `did` (`DidUri`) which specifies the DID that must sign the data
+
+The call back is expected to return `SignResponseData` which contains
```
ntn-x2:
```suggestion
The callback is expected to return `SignResponseData` which contains
```

docs/develop/01_sdk/02_cookbook/06_signCallback.md
ntn-x2: I guess also using the TS would work, perhaps.

docs/develop/01_sdk/02_cookbook/06_signCallback.md
```diff
@@ -0,0 +1,78 @@
...
+
+The `SignExtrinsicCallback` is called with the same `SignRequestData`, but can return `SignResponseData` that doesn't contain the `keyUri`.
+
+<TsJsBlock>
+    {SignExtrinsicCallback}
+</TsJsBlock>
+
+### GetStoreTxSignCallback
+
+The `GetStoreTxSignCallback` is only used to sign the data that is submitted to the blockchain when a DID is registered.
```
ntn-x2:
```suggestion
The `GetStoreTxSignCallback` is only used to sign the data that is submitted to the blockchain when a DID is being created.
```

docs/develop/01_sdk/02_cookbook/06_signCallback.md
```diff
@@ -0,0 +1,78 @@
...
+Because there is no DID identifier before the DID is registered on chain, this callback doesn't receive the DID as a parameter.
+There is also no DID document and no public key stored if the DID isn't registered yet.
```
ntn-x2:
```suggestion
There is also no DID document and no public key stored if the DID hasn't yet been created.
```

docs/develop/01_sdk/02_cookbook/06_signCallback.md
```diff
@@ -0,0 +1,78 @@
...
+Therefore the `keyUri` can not point to a valid DID key.
```
ntn-x2:
```suggestion
Therefore the `keyUri` cannot point to a valid DID key.
```

code_examples/core_features/signCallback/signExtrinsicCallback.ts
```diff
@@ -0,0 +1,19 @@
+import * as Kilt from '@kiltprotocol/sdk-js'
+import { lookupDidDocument, lookupDidKeyPair } from '.'
```
ntn-x2: This should not show up in the user-visible code, so it should perhaps be wrapped in a TsJsSnippet, and not a TsJsBlock, so that we can hide these imports. Alternatively, the functions should be declared in the file, as otherwise the user has not way to know what these functions do.

code_examples/core_features/signCallback/signCallback.ts
```diff
@@ -0,0 +1,19 @@
+import * as Kilt from '@kiltprotocol/sdk-js'
+import { lookupDidDocument, lookupDidKeyPair } from '.'
```
ntn-x2: Same here.

code_examples/core_features/signCallback/getStoreTxSignCallback.ts
ntn-x2: We should strive to only import the Kilt package as such, as we have done that throughout the docs and I think it makes it easier for users to read, as they don't have to know from where to import stuff.

code_examples/core_features/run_core_features.ts
ntn-x2: @arty-name what's your opinion about this?

ntn-x2 (commented):
I quickly went over the changes, and have few things to point out that I think would improve code readability and keep the style consistent with how the docs are currently structured. If you have decided otherwise, that is fine since I was not involved in the discussion, but I think we can do better when writing these code snippets.
P.S. Very good section to have, and very much needed 👌

docs/develop/01_sdk/02_cookbook/06_signCallback.md
weichweich: You want to add the call signature? I thought about that as well. I wanted to export the type definition from the SDK and even had the crazy idea to add the SDK as a git submodule or similar. 😁

docs/develop/01_sdk/02_cookbook/06_signCallback.md
weichweich: [9e73861](https://github.com/KILTprotocol/docs/pull/181/commits/9e738613d5307cdd06ee60bd310bf4505b7b36a3)

code_examples/core_features/signCallback/signExtrinsicCallback.ts
weichweich: I think it's fine to show up. It needs to be clear where these methods come from. I think it's not clear enough that the user has to implement them themself.
I tried to make it clearer by explaining everything step by step. But i would also maybe add a disclaimer that these methods need to be implemented by the dev or an extension.

[0036481](https://github.com/KILTprotocol/docs/pull/181/commits/0036481e0266fdbdc98c3afb20b807a2f8bbb2e5)

code_examples/core_features/run_core_features.ts
arty-name: One thing that is definitely lost in this change is the re-throwing of errors. I assume it might have been important for detection of failures, in that case it should be restored. Let’s see how the code looks to you afterwards. Personally, I always prefer async/await to promises, also because with them you do not randomly lose re-throwing of errors ;)

arty-name:
So, I read the documentation, and I, personally, think that Sporran takes a simpler approach by creating multiple DID-bound callbacks instead of attempting a universal one. With that approach, we have no need to look up documents or keypairs, and the callbacks are quite straightforward. Examples of this simplicity can even be seen in this repository in `code_examples/core_features/claiming/index.ts`. I believe advocating for this approach will make integrating SDK simpler for external developers.

code_examples/core_features/run_core_features.ts
weichweich: We can also re-throw the errors with the `catch` variant. Wasn't sure what the logic was there.

I think the `catch` variant is less verbose and more readable and @Dudleyneedham seems to agree. But if the two of you disagree, we can keep the status quo.

[58d4175](https://github.com/KILTprotocol/docs/pull/181/commits/58d417596c4eb9b6dd59ddded29dd7cc254ddbb3)

arty-name: If the readability is too bad, the logging could be moved inside the functions themselves, I believe. This piece of code running things in parallel and logging at the same time is not particularly elegant in any form, I’d say.

weichweich:
> by creating multiple DID-bound callbacks instead of attempting a universal one

Yes I would love to take that approach as well. But looking at this problem from a dApp developer and not from an extension developer makes it become much more complex. We suddenly only know the DID identifier. So we can't used use the key pair from the site where we create the call back. The key pair is not known to the dApp. We rather would need to defer the call to an extension or to some key store that safely stores the private keys. The examples where the most generic that i could come up with.
I also want to discourage storing the secret key in your dApp rather than the extension.

With the current design of the extension api the call back would look like that:

```

export const signCallback: Kilt.SignCallback = async ({
  data,
  keyRelationship,
  did: didUri
}) => {
  const extension = findExtensionThatStoresDid(didUri)
  while (true) {
    const signedResponse = extension.requestSignature(data)
    if (signature.did === didUri && signature.keyRelationship === keyRelationship) {
      return signedResponse
    } else {
      // show error to user and try again
    }
  }
}

```

Ok I don't know the exact API but the important part is the while loop until the user used the correct DID. Having this in the docs would really make it easy for the developers. We actually wouldn't even require them to implement that, we would just provide the callbacks as a library function, ready to use for them.

arty-name:
You said "dApp developers" and that gave me a new perspective on the topic. Yep, it’s totally fair from this viewpoint! Would it make sense though to have both approaches documented? First a happy path "you control the keys" and then the more flexible but complex approach for the dApp developers "you do not control the keys"?

weichweich:
> First a happy path "you control the keys" and then the more flexible but complex approach for the dApp developers "you do not control the keys"?

I think if we cover these two cases we should be good. I'll change that then.

docs/develop/01_sdk/02_cookbook/06_signCallback.md
```diff
@@ -0,0 +1,80 @@
...
+May it be a ledger, an air gapped phone or your browser extension.
+The interface is generic enough to support implementations for all these security measures.
+
+## The SignCallback Family
+
+There are three types of signing callbacks.
+The `SignCallback` is the most general and can be used in almost all cases, except when signing a full DID creation transaction.
+The `SignExtrinsicCallback` is a special `SignCallback` which can only be used to sign extrinsics.
+Every `SignCallback` can also be used as a `SignExtrinsicCallback`.
+The `GetStoreTxSignCallback` can only be used to sign the creation of a new DID.
```
wischli: C: Use (number) list there

```suggestion
There are three types of signing callbacks:
1. The `SignCallback` is the most general and can be used in almost all cases, except when signing a full DID creation transaction.
2. The `SignExtrinsicCallback` is a special `SignCallback` which can only be used to sign extrinsics. 
Thus, every `SignCallback` can also be used as a `SignExtrinsicCallback`.
3. The `GetStoreTxSignCallback` can only be used to sign the creation of a new DID.
```

docs/develop/01_sdk/02_cookbook/06_signCallback.md
```diff
@@ -0,0 +1,80 @@
...
+
+### SignCallback
+
+The plain `SignCallback` signs arbitrary data.
+It is called with `SignRequestData` which contains
+
+* the `data` (UInt8Array) that should be signed
```
wischli: S: Codify `UInt8Array` as we use it like that a few lines below.
```suggestion
* the `data`  as `UInt8Array` that should be signed
```

docs/develop/01_sdk/02_cookbook/06_signCallback.md
```diff
@@ -0,0 +1,80 @@
...
+</TsJsBlock>
+
+### SignExtrinsicCallback
+
+The `SignExtrinsicCallback` is a special case of the `SignCallback`.
+Signing an extrinsic doesn't require the `keyUri` as a return value since the chain will pick the appropriate key using information from the extrinsic.
+The extrinsic that is submitted has a specific `VerificationKeyRelationship`, which defines which key must be used to sign the extrinsic.
+Using this relation between extrinsic and key, the chain looks up the public key and verifies the signature.
+
+The `SignExtrinsicCallback` is called with the same `SignRequestData`, but can return a `SignResponseData` that doesn't contain the `keyUri`.
```
wischli: S: To me, it's confusing that above we state three properties of the `SignResponseData` and here refer to it but without the `keyUri`. As a normal/new user, I would have expected `SignResponse` to always include the three above fields. 

I think, just to be clear, we should should have bullet points describing the return type for every callback type, e.g.

```suggestion
The `SignExtrinsicCallback` is called with the same `SignRequestData`, but can return a `SignResponseData` that doesn't contain the `keyUri` but only

* the `signature` as an `UInt8Array`
* and the `keyType` which specifies the signature scheme that was used (either `sr25519`, `ed25519` or `ecdsa`)
.
```

docs/develop/01_sdk/02_cookbook/06_signCallback.md
```diff
@@ -0,0 +1,80 @@
...
+    {SignExtrinsicCallback}
+</TsJsBlock>
+
+### GetStoreTxSignCallback
+
+The `GetStoreTxSignCallback` is only used to sign the data that is submitted to the blockchain when a DID is being created.
+Because there is no DID identifier before the DID is registered on chain, this callback doesn't receive the DID as a parameter.
+There is also no DID document and no public key stored if the DID hasn't yet been created.
+Therefore the `keyUri` cannot point to a valid DID key and is not included in the return data.
+
```
wischli, line 76: As above, I would add bullet points for the `SignRequestData` and `SignResponseData` for this type of callback. I know it's a repetition, but I see lots of users getting confused here. In the end, documentation should be crystal clear and super easy to follow.

wischli (approved):
Overall LGTM! I would add some more explanation, see below. But maybe the code examples are good enough.
Of course, I am not the best person to review the code snippets but I couldn't find flaws there.

PR #194, opened 2023-01-04
fix: move to polkadot

weichweich:
## fixes https://github.com/KILTprotocol/ticket/issues/2352

Fix commands that still use Kusama as a relaychain.

## Checklist:


Dudleyneedham (approved):
LGTM

PR #196, opened 2023-01-04
fix: dapp guide

weichweich:
## fixes https://github.com/KILTprotocol/ticket/issues/2358

This PR fixes the well-known DID configuration guide.

* The attestation step contains multiple errors (missing DID authorization, wrong parameter in the call)
* The order of the guide leads the reader to first fail in the first step, only to find out they should have done the second step first
* the guide is hard to follow in general
* no explanation is given why the well-known DID configuration is needed

## Checklist:


gabrieljaegerde:
Thanks for getting on top of this. There is currently a jump in the docs in [​Formatting the Domain Linkage Credential](https://pr-196.d3oeht7cvddn19.amplifyapp.com/docs/develop/dApp/well-known-did-config#formatting-the-domain-linkage-credential) step. Specifically it would be great to show where the `domainLinkageCredential` comes from in that code snippet.

Dudleyneedham:
We do not need to store the DID configuration on-chain. This section can be removed.

weichweich:
I reverted all the change unrelated to the content of the guide. This includes the tests for the code, but since I didn't change the code after the last successful test, everything should still work.

docs/develop/07_dApp/02_well-known-did-config.md
```diff
@@ -0,0 +1,109 @@
...
+
+
+:::danger This is a working draft
+
+The KILT support of the *Well-Known DID Configuration* uses unpublished specifications and will change in the future.
+
+:::
+
+The *Well-Known DID Configuration* is implemented as a security measure when setting up the communication session between the dapp and extension.
+It ensures that the DID that the browser extension is communicating to is linked to the domain that is visited by the browser.
```
ntn-x2:
```suggestion
It ensures that the DID the browser extension is communicating to is linked to the domain that is visited by the browser.
```

docs/develop/07_dApp/02_well-known-did-config.md
```diff
@@ -0,0 +1,109 @@
...
+This rule is currently enforced by the KILT DID Wallet reference implementation (Sporran Extension), but might be relaxed in the future.
```
ntn-x2:
```suggestion
This rule is currently enforced by the KILT Wallet reference implementation (Sporran Extension), but might be relaxed in the future.
```

Is it a DID Wallet? Right now I think it's just a wallet.

docs/develop/07_dApp/02_well-known-did-config.md
```diff
@@ -0,0 +1,109 @@
...
+The implementation is based on the [*Well-Known DID Configuration*](https://identity.foundation/.well-known/resources/did-configuration/) specified by the Decentralized Identity Foundation.
+
+Once a communication session between a dapp and an extension is opened, the extension will query `<domain-name>/.well-known/did-configuration.json`.
+This JSON-file must contain a credential presentation that conforms to the [Domain Linkage CType][CType-Domain-Linkage].
+
+
+## Setup the Well-Known DID Configuration
```
ntn-x2:
```suggestion
## Set up the Well-Known DID Configuration
```

docs/develop/07_dApp/02_well-known-did-config.md
```diff
@@ -0,0 +1,109 @@
...
+    * The `authenticationKey` is required for signing the transaction
+1. Create a claim
+2. Attest the claim
+3. Create a presentation
+4. Host the presentation on your website at `https://<your domain>/.well-known/did-configuration.json`
+
+### Create a DID
+
+Your dapp needs a DID to identify itself to the extension.
+If you don't have a DID yet, follow the [*create a full DID* guide](../01_sdk/02_cookbook/01_dids/02_full_did_creation.md).
```
ntn-x2:
```suggestion
If your dapp does not have a DID yet, follow the [*create a full DID* guide](../01_sdk/02_cookbook/01_dids/02_full_did_creation.md).
```

docs/develop/07_dApp/02_well-known-did-config.md
```diff
@@ -0,0 +1,109 @@
...
+Make sure to create your DID with an `assertionMethodKey` so that you are able to issue attestations.
```
ntn-x2:
```suggestion
Make sure to create the DID with an `assertionMethodKey` so that you are able to issue attestations.
```

docs/develop/07_dApp/02_well-known-did-config.md
```diff
@@ -0,0 +1,109 @@
...
+
+### Making the claim
+
+After you got a DID, we can make a claim about that DID.
```
ntn-x2:
```suggestion
After you get a DID, you can make a claim about that DID.
```

You? We? Your dapp? Pronouns should be made consistent.

docs/develop/07_dApp/02_well-known-did-config.md
```diff
@@ -0,0 +1,109 @@
...
+<TsJsSnippet>
+  {DomainLinkageClaim}
+</TsJsSnippet>
+
+The credential isn't attested yet and is therefore not valid yet.
+
+### Self-attesting the credential
+
+A valid credential requires an attestation.
+Since we want to signal that the DID whats to link itself to the website, the DID has to attest to the claim itself.
```
ntn-x2:
```suggestion
Since the website wants to link itself to the DID just created, it has to self-attest the domain linkage credential, i.e., write the credential attestation on chain using the same DID it is trying to link to.
```

The DID does not attest anything, since it's just an identifier. I tried to rephrase the sentence a bit to make it easier to read and also more technically correct.

docs/develop/07_dApp/02_well-known-did-config.md
```diff
@@ -0,0 +1,109 @@
...
+
+<TsJsSnippet>
+  {AttestCredential}
+</TsJsSnippet>
+
+If you want to learn more about attestations you can refer to our [concept guide](../../concepts/04_credentials/04_attestation.md) or the [cookbook](../01_sdk/02_cookbook/04_claiming/03_attestation_creation.md)
+
+### Presenting the credential
+
+Before we can host the credential on the website, we need to derive a presentation from it.
```
ntn-x2:
```suggestion
To use the newly attested credential, we need to derive a presentation from it to host on the dapp website.
```

We host the presentation and not the credential, right?

docs/develop/07_dApp/02_well-known-did-config.md
```diff
@@ -0,0 +1,109 @@
...
+
+<TsJsSnippet>
+  {SignPresentation}
+</TsJsSnippet>
+
+
+The Well Known DID Configuration specification requires a specific format (JSON-LD) for the domain linkage credential.
```
ntn-x2: 1. Well Known, or Well-known?
2. Can we just say the credential needs to be a verifiable credential, linking to the spec? That implies it's a JSON-LD document.

docs/develop/07_dApp/02_well-known-did-config.md
```diff
@@ -0,0 +1,109 @@
...
+For now we have to manually convert our credential into the required format.
```
ntn-x2:
```suggestion
For now we have to manually convert our KILT credential into the required format.
```

docs/develop/07_dApp/02_well-known-did-config.md
```diff
@@ -0,0 +1,109 @@
...
+### Host the Presentation
+
+Now that you generated a presentation, you need to host it in your web app, so that the extension can query the presentation.
+The extension will make an HTTP GET request to the following URI, and your dapp must respond with the presentation.
+
+`/.well-known/did-configuration.json`
+
+How the file is hosted depends on your project setup and is out of scope for this guide.
+
+[CType-Domain-Linkage]: https://github.com/KILTprotocol/ctype-index/tree/main/ctypes/0x9d271c790775ee831352291f01c5d04c7979713a5896dcf5e81708184cc5c643
```
ntn-x2, line 113: Some links are collected here, some are not. Maybe we can collect them all here?

code_examples/dapp/src/dapp/signPresentation.ts
```diff
@@ -1,38 +1,32 @@
...
-    }
+export async function main({
+  didUri,
+  assertionMethodKey,
+  domainLinkageCredential
+}: Parameter) {
+  // We need the KeyId of the AssertionMethod Key. There is only
+  // one AssertionMethodKey and its id is stored on the blockchain.
+  const didResolveResult = await Kilt.Did.resolve(didUri)
+  if (typeof didResolveResult.document === 'undefined') {
```
ntn-x2, line 17:
```suggestion
  if (!didResolveResult.document) {
```

I think this is the way we check for undefined values.

docs/develop/07_dApp/02_well-known-did-config.md
```diff
@@ -0,0 +1,109 @@
...
+After you got a DID, we can make a claim about that DID.
+The claim has to be based on the [Domain Linkage CType][CType-Domain-Linkage]:
+
+<TsJsBlock>
+  {DomainLinkageCtype}
+</TsJsBlock>
+
+The credential is built from the CType, claim contents, and your dapp's unique DID:
+
+<TsJsSnippet>
```
ntn-x2: Did you mean to use the snippet? I think it is not ideal that every snippet now has a return statement. Either we hide the return line, or we show the whole function including its signature.

ntn-x2 (commented):
Few comments, mostly really minor stuff. The main point for me is to keep a bit of consistency regarding the usage of 'we', 'you', 'the dapp', 'the website', etc.

code_examples/dapp/src/dapp/signPresentation.ts
weichweich: The reason for doing this `typeof` stuff is that this is the way to only check for undefined. The `!` also returns `true` in the following cases:

* ""
* 0
* null
* false
* NaN

And since undefined could be overwritten with something else, `typeof` is used instead of comparing.

docs/develop/07_dApp/02_well-known-did-config.md
weichweich: yeah was also not super happy about that. But even though it's not pretty, it doesn't hurt readability. It shows what's the outcome of the snippet. So it should even help to understand the flow of the script better.

we can show the complete file if you want?

docs/develop/07_dApp/02_well-known-did-config.md
weichweich: I think it's well-known what's correct 2f725b7c2f728cb053e43b6275f09671f5a64341

docs/develop/07_dApp/02_well-known-did-config.md
weichweich: 0d62d5fcbc8cc956a9f969ab5665eefb2d6daa11

docs/develop/07_dApp/02_well-known-did-config.md
ntn-x2: I wouldn't show the complete block since the `main` function takes some parameters which we don't necessarily want to show to the users. For me, the choice is either between the way things are now, or by removing the return line (which is possible). I prefer the latter, but I leave the final decision to you, as I don't have a strong opinion.

code_examples/dapp/src/dapp/domainLinkageClaim.ts
```diff
@@ -1,19 +1,19 @@
...
-
+export type Parameter = {
+  domainLinkageCType: Kilt.ICType
+  didUri: Kilt.DidUri
+}
+export function main({ domainLinkageCType, didUri }: Parameter) {
   const claimContents: Kilt.IClaimContents = {
-    id: did,
-    origin: 'https://example-dApp.com'
+    id: didUri,
```
rflechtner, line 10: that's the same as the claim owner though, right? do we need it?

rflechtner: btw remember the VC discussion? Naming a claim `id` means we set ourselves up for disaster - `type` and `id` are reserved in that context

rflechtner (commented):
Two questions:
1. Why do we have the `id` property on the CType which simply contains the claim owner?
2. Do we still format the credential as a VC? It reads like it but you explained it differently yesterday.

docs/develop/07_dApp/02_well-known-did-config.md
weichweich: [7907375](https://github.com/KILTprotocol/docs/pull/196/commits/7907375717ba778338261db1125faab29f31fb51)

code_examples/dapp/src/dapp/domainLinkageClaim.ts
weichweich: These two are the required fields by the w3c spec. The domain linkage credential has `id` and `origin` fields.
For this PR we leave it as it is. This is only the improvement for the current strategy. Next step is to update this guide to the simple solution and in the future to have the spec complaint solution.

rflechtner: I think there is a misunderstanding here; credentialSubject.id is not a claim, it's the credential subject's identifier, i.e. the claim owner. When we export KILT credentials to VCs, the credential owner becomes the credentialSubject.id. If we tried to export a claim with this CType we'd be adding the id twice.

ntn-x2: Indeed. This slipped through my review. The `id` property is actually the KILT credential subject. So the only claim in the credential is actually the `origin` property.

weichweich: The [CType](https://github.com/KILTprotocol/ctype-index/blob/main/ctypes/0x9d271c790775ee831352291f01c5d04c7979713a5896dcf5e81708184cc5c643/ctype.json) was created with both `origin` and `id`. So the claim has to have both as well.

ntn-x2: And we fix this inconsistency afterwards? That would also work, not to break stuff right away.

weichweich: yeah i wouldn't handle that now

rflechtner:
There's a general issue with kilt attestations and the well-known did configuration specifications, which revolve around this section: https://identity.foundation/.well-known/resources/did-configuration/#domain-linkage-credential

A domain linkage credential is described as follows:
> - id MUST NOT be present.
> - issuanceDate MUST be present.
> - expirationDate MUST be present.
> - credentialSubject.id MUST be present, and MUST be a [DID](https://identity.foundation/.well-known/resources/did-configuration/#DID).
> - credentialSubject.origin MUST be present, and MUST be a domain [Origin](https://identity.foundation/.well-known/resources/did-configuration/#Origin).

As I mentioned in [this summary on VCs and KILT credentials](https://github.com/KILTprotocol/spec-KiltCredentialV1/pull/1#issuecomment-1375841562) we have a hard time already to secure an `issuanceDate`;
securing an `expirationDate` property is off the table completely (if we don't put it on the credentialSubject instead).

So if we want to be really compliant with what's described in the specs, we'd probably have to cryptographically sign a VC or a JWT just as described in the specifications.
We could even do that, I assume, if we just specify that the assertionMethod key must be of a type that is supported by existing JWT/linked data signature schemes (ECDSA or ed25519 I would think).

ntn-x2 (approved):
Looks good to me!

PR #198, opened 2023-01-11
fix: wrong naming

weichweich:
## Checklist:

### DIFF
--- a/code_examples/dapp/src/dapp/sessionCheck.ts
+++ b/code_examples/dapp/src/dapp/sessionCheck.ts
@@ -3,7 +3,7 @@ import { u8aToHex } from '@polkadot/util'
 import * as Kilt from '@kiltprotocol/sdk-js'
 
 let session: {
-  encryptionKeyUri: Kilt.DidResourceUri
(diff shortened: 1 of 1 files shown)


code_examples/dapp/src/dapp/sessionCheck.ts
```diff
@@ -3,16 +3,16 @@ import { u8aToHex } from '@polkadot/util'
 import * as Kilt from '@kiltprotocol/sdk-js'
 
 let session: {
-  encryptionKeyUri: Kilt.DidResourceUri
```
ntn-x2, line 6: If it's a `DidResourceUri` then `keyUri` was the right name. Why has this been renamed?

weichweich: Yes that is right. I blindly changed it to what sporran does.

But the spec says it's calls a URI

weichweich: https://github.com/KILTprotocol/ticket/issues/2380

It's an issue with the versions. The spec version is 3 while this guide was never updated to the latest spec version

PR #218, opened 2023-02-17
fix: concurrency issues by having four different test accounts

ntn-x2:
Replace the current approach of starting all test flows from the faucet account with a different flow where four sub-accounts are derived from a single seed.
The accounts are already funded on Peregrine, and are derived with the paths `//core`, `//workshop`, `//dapp`, and `//staking` respectively. 

This solves the annoying concurrency issues that we've been running into multiple times now.

### DIFF
--- a/code_examples/sdk_examples/src/test.ts
+++ b/code_examples/sdk_examples/src/test.ts
@@ -68,7 +68,9 @@ const MNEMONIC_ENV = 'BASE_MNEMONIC'
     throw new Error('Account seed is missing')
   }
 
-  const baseAccount = await new Keyring({ type: 'sr25519' }).addFromMnemonic(mnemonic)
+  const baseAccount = await new Keyring({ type: 'sr25519' }).addFromMnemonic(
+    mnemonic
+  )
 
   // If any of these flows fail, just send some more tokens to the account that is failing.
   try {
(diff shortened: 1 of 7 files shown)


code_examples/sdk_examples/src/test.ts
```diff
@@ -56,31 +58,33 @@ const SEED_ENV = 'FAUCET_SEED'
...
+      `The base mnemonic to generate the test accounts has not been specified.
+        Set the secret seed using the ${MNEMONIC_ENV} environment variable.`
     )
     throw new Error('Account seed is missing')
   }
-  const faucetAccount = Kilt.Utils.Crypto.makeKeypairFromSeed(
-    hexToU8a(faucetSeed),
-    'sr25519'
+
+  const baseAccount = await new Keyring({ type: 'sr25519' }).addFromMnemonic(
```
Dudleyneedham, line 71: I prefer this way.

code_examples/sdk_examples/src/test.ts
```diff
@@ -56,31 +58,33 @@ const SEED_ENV = 'FAUCET_SEED'
...
+    mnemonic
   )
 
+  // If any of these flows fail, just send some more tokens to the account that is failing.
   try {
     if (whichToRun.workshop) {
-      await testWorkshop(faucetAccount, wssAddress)
+      await testWorkshop(baseAccount.derive('//workshop'), wssAddress)
```
Dudleyneedham, line 78: This makes sense.

Dudleyneedham: How do they get funding?

Dudleyneedham (commented):
Looks good. Only one minor comment. How do each of the derived accounts get funding?

Dudleyneedham (approved):
LGTM

PR #221, opened 2023-02-23
chore: add faucet retro compatibility

ntn-x2:
Add support for old way of using the faucet, for retro-compatibility when running against a local standalone dev deployment.


weichweich (approved):
GLTM

PR #224, opened 2023-03-06
fix: links and more

weichweich:
## fixes no ticket

* Don't check links while building. We have a pipeline for that. It would be nice to have a preview even if links are broken.
* Fix broken link
* remove duplicate file
* fix extrinsic name in delegate guide

## Checklist:


ntn-x2 (commented):
I see that `check-links` is run by the CI, but what about `check-ts`? I personally like to call `build` locally before I push anything on the CI, so perhaps instead of changing that, we can add a new script `build:docs` which just generates the examples and calls docusaurus?

weichweich:
> but what about `check-ts`?

Ah to be honest, I just thought was `check-links` twice when I removed that. Didn't notice that there was a `check-ts` instead. 😳 
I added both back. Maybe it's good that they are a bit annoying, so that nobody skips these checks. 😁

trusch (approved):

ntn-x2 (approved):

PR #232, opened 2023-03-30
feat: add search bar

ntn-x2:
Fixes https://github.com/KILTprotocol/ticket/issues/2413.

Docusaurus guide at https://docusaurus.io/docs/2.2.0/search#using-algolia-docsearch.

Not sure there's an easy way to test this before it gets merged, since the index is tied to the official `https://docs.kilt.io` domain name. I would suggest to merge this, wait for the first crawl to happen, and then to fix any issues.

### DIFF
--- a/docusaurus.config.js
+++ b/docusaurus.config.js
@@ -15,7 +15,25 @@ module.exports = {
     mermaid: true,
   },
   themes: ['@docusaurus/theme-mermaid'],
+  // Config taken from https://docusaurus.io/docs/search#using-algolia-docsearch
+  // `appId`, `apiKey`, and `indexName` were provided in the email
   themeConfig: {
+    algolia: {
+      // The application ID provided by Algolia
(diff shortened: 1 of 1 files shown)


rompemoldes:
Is it possible to test it locally? Should I clone the docs repository?  If not, then I would agree with Antonio, just merge first and than test it. 

On Amplify it looks good, but it does not work well.

ntn-x2:
@kilted-andres when requesting access to Algolia, they asked for the website URL. I suspect they crawl and index only that one. Hence, I am afraid it's not possible to test the search functionalities neither locally, nor in Amplify.

weichweich (approved):

rompemoldes:
The merged search bar is not working. Querying "DID" shows no results. See attached picture: 
SearchBar failed

ntn-x2:
Yes I noticed that. I will look into that.

PR #235, opened 2023-04-06
fix: paths

ntn-x2:


rflechtner (approved):
LGTM - Probably unrelated, but I don't seem to be able to build this locally though, it fails with `Cannot find module '@babel/helper-get-function-arity'`

ntn-x2:
Could it be because of a different Node version? On my machine and on the CI it compiles fine. We don't really have a requirement here.

rflechtner:
> Could it be because of a different Node version? On my machine and on the CI it compiles fine. We don't really have a requirement here.

We have an .nvmrc which I used, but I can check what the ci uses

PR #249, opened 2023-05-31
deps: upgrade to sdk v0.33.0

### LINKED ISSUE #248: [AUTOMATIC] New SDK 0.33.0-rc.1 is incompatible with deployed docs examples
## Incompatible SDK detected

[At least one of the test workflows](https://github.com/KILTprotocol/docs/actions/runs/5121659326) using the latest released SDK 0.33.0-rc.1 on docs has failed.

### Tasks

- [ ] Check if any docs need to be updated, and if needed do so

rflechtner:
## Checklist:

    * Link the PR/Ticket here

### DIFF
--- a/code_examples/sdk_examples/package.json
+++ b/code_examples/sdk_examples/package.json
@@ -13,7 +13,7 @@
     "test": "ts-node src/test.ts"
   },
   "dependencies": {
-    "@kiltprotocol/sdk-js": "0.33.0-rc.3",
+    "@kiltprotocol/sdk-js": "0.33.0",
     "axios": "^0.27.2",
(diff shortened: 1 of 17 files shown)


arty-name:
Should we also remove from docs the mentions of polyfills?

arty-name (approved):
The topics I’m familiar with look good

rflechtner:
> Should we also remove from docs the mentions of polyfills?

We are confident these are not needed any more right? In that case I agree we should

Edit: this seems like it may change a lot of recommendations we make, so I'd open a ticket for that (https://github.com/KILTprotocol/ticket/issues/2742)

PR #251, opened 2023-06-08
refactor: remove verify_attestation snippet from quickstart

rflechtner:
This is an oddity in the current Quickstart; we verify the attestation twice. I don't think that's helpful in this format.

## Checklist:

    * Link the PR/Ticket here


arty-name (approved):

arty-name:
The numbering is now off: it jumps from 06 straight to 08

rflechtner:
> The numbering is now off: it jumps from 06 straight to 08

I was wondering if that's an issue ?

EDIT: you're right, it's ugly, I fixed it

PR #252, opened 2023-06-12
refactor: deposit and MAC

Ad96el:
## fixes [#2707](https://github.com/KILTprotocol/ticket/issues/2707) and [#2605](https://github.com/KILTprotocol/ticket/issues/2605)

It is explained how high the deposits and fees are for the different blockchain entities. 

In practice, the MAC is often called nonce ⇾ update it in the message section.


docs/concepts/07_messaging.md
```diff
@@ -25,7 +25,7 @@ In order to send a message to the other party, a DID owner (let's call her **Ali
 Using this key in combination with her secret encryption key, **Alice** can now encrypt the message such that only she and **Bob** can decrypt it.
 
 **Bob** can decrypt this message after looking up **Alice's** encryption key.
-An additional _message authentication code_ (MAC) added during encryption and verified on decryption protects against manipulation of the encrypted data.
+An additional _message authentication code_ (MAC or in practice often called nonce) added during encryption and verified on decryption protects against manipulation of the encrypted data.
```
rflechtner: The nonce that is required for libsodiums `box` algorithm, if I got that right, is used to add entropy in an x25519 Diffie-Hellmann key agreement to make sure a new shared key is generated each time. This key is then used in a salsa20 stream cipher to encrypt the content and then for the generation of an additional Poly1305 MAC to protect against bit flipping attacks, which would allow an attacker to manipulate the content of the message. In short: MAC and nonce, not the same.

The nonce is input to the encryption algorithm; to keep its security guarantees intact, each time you encrypt a message you need to use a different nonce. This can be achieved by using a random value from a cryptographically secure source of randomness. When encrypting a message using the kilt sdk, that's what happens.

You need the value of the nonce used for encryption to decrypt the message again, because a different nonce would result in a different shared key, which would not be able to decrypt the ciphertext or verify the MAC.

rflechtner (changes requested):
I like the deposit descriptions! The explanation of the nonce used in message encryption is flawed though. Hope my explanation helps?

docs/concepts/07_messaging.md
Ad96el: I see. It should be resolved [here](https://github.com/KILTprotocol/docs/pull/252/commits/b5964743b5c495b719f70ccc0f7a3908f3ed1aaf). I only explained the conceptional usage of a nonce. I hope it is fine now :pray:.

rflechtner: nice, that reads great! good explanation

rflechtner (approved):
🚀

PR #254, opened 2023-06-14
fix: adding the uri to the env

Dudleyneedham:
## fixes

Fixes the environment return on generateDID for Attester as it is required for the verification but not logged out.

## Checklist:

    * Link the PR/Ticket here


code_examples/sdk_examples/src/workshop/attester/generateDid.ts
```diff
@@ -61,10 +61,11 @@ if (require.main === module) {
       // Load attester account
       const accountMnemonic = process.env.ATTESTER_ACCOUNT_MNEMONIC as string
       const { account } = generateAccount(accountMnemonic)
-      const { mnemonic } = await createFullDid(account)
+      const { mnemonic, fullDid } = await createFullDid(account)
 
       console.log('\nsave following to .env to continue\n')
       console.error(`ATTESTER_DID_MNEMONIC="${mnemonic}"\n`)
+      console.error(`ATTESTER_DID_URI="${fullDid.uri}"\n`)
```
rflechtner, line 68: this should be mentioned in the text (i.e. in the markdown) as well, along with the mnemonic

Dudleyneedham: Agreed. I have updated

rflechtner: I was talking specifically about line 93-95 in the file below, that should be updated to reflect that we've got two env variables being logged to console now

Dudleyneedham: Understood :)

rflechtner (approved):
looks good to me now!

PR #260, opened 2023-07-13
feat: Public Creation removal of CType Creation

Dudleyneedham:
## fixes https://github.com/KILTprotocol/ticket/issues/2703

Adding additional burden onto the docs user to build a CType was unnecessary. Removal of the CType takes away some of the friction.

Using existing CType will make this easier, though a reference to CType creation was added.

## How to test:
yarn test (DUH)

## Checklist:

    * Link the PR/Ticket here


ntn-x2:
Point is, I think when I worked on this snippet, there were requests to show the full flow, so people would not have to click right and left. Is this what we want here? That someone wanting to create a public credential now has to click and follow the guide on how to create a CType? It's fine by me, but back then people had different ideas.

Dudleyneedham:
> Point is, I think when I worked on this snippet, there were requests to show the full flow, so people would not have to click right and left. Is this what we want here? That someone wanting to create a public credential now has to click and follow the guide on how to create a CType? It's fine by me, but back then people had different ideas.

I believe that we are trying to reduce the burden on the user who wants to try public credentials. I think developing on the ideas and reasoning from last time: Yes the problem of making the user switch isn't great. But I think we should condense these ideas. I am open to ideas.

weichweich:
> on this snippet

IMO a snippet is a very small part of code that does exactly one thing.

What you are talking about is a tutorial that explains a common workflow with public credentials from start to finish.

I think both have their right to exist.

Dudleyneedham:
This is an overall issue with our thought process. I think Albi is right, and we should maybe define each of these as a tutorial or an example. Therefore, we can be more aligned. 

I do think we would need a tutorial for most things. This can utilise more in-depth end-to-end explanations and practical examples for the user to see the whole flow.

weichweich:
I thought about the cookbook more as a collection of snippets, but it's probably rather a collection of how-to guides. So each of the entries should be self contained and reproducible without additional setup.

ntn-x2:
Should this be discussed as part of a larger documentation strategy meeting?

ntn-x2:
I feel we are not yet all aligned on how the docs should look like.

Dudleyneedham:
I think so too. But the main outcome of our meeting was to leave it for the tech writer. We only do the necessary docs upgrades and decisions. 

That being said, I think I will bring it up in the daily.

code_examples/sdk_examples/src/core_features/public_credentials/01_create_credential.ts
```diff
@@ -1,10 +1,28 @@
...
+        type: 'integer'
+      },
+      creationDate: {
+        type: 'string'
+      },
+      artistIdentity: {
+        type: 'string'
+      }
+    }
+  )
```
weichweich: Move this out of the function. That makes it more readable.

docs/develop/01_sdk/02_cookbook/05_public_credentials/01_credential_issuance.md
```diff
@@ -5,28 +5,40 @@ title: Credential Issuance
...
+        type: 'integer'
+      },
+      creationDate: {
+        type: 'string'
+      },
+      artistIdentity: {
+        type: 'string'
+      }
+    }
+  )
```
weichweich: Remove this. It's untested and a duplication of the code below

weichweich (changes requested):
I would not show the `CType.fromProperties` twice. DRY principle :D

docs/develop/01_sdk/02_cookbook/05_public_credentials/01_credential_issuance.md
Dudleyneedham: This has been removed.

code_examples/sdk_examples/src/core_features/public_credentials/01_create_credential.ts
Dudleyneedham: Done

weichweich (approved):

PR #274, opened 2023-08-25
feat: updating the official docs to include extension library

Dudleyneedham:
## fixes https://github.com/KILTprotocol/ticket/issues/2822

Moving the extension api docs and read me to the official documentation

## How to test:

Due to the extension library being unreleased it can't be officially tested.

## Checklist:

    * Link the PR/Ticket here


weichweich (commented):
I would wait until we have an official release. Before that, it doesn't make sense to document it, since it's unusable.

Dudleyneedham:
Agreed. I think we should wait

ChrisChinchilla:
I renamed the branch to match ticket names, so this wasn't closed, but just moved…

PR #293, opened 2024-01-16
Update Docusaurus and related dependencies

ChrisChinchilla:
## fixes KILTProtocol/ticket#3028

### DIFF
--- a/src/components/SnippetBlock/index.js
+++ b/src/components/SnippetBlock/index.js
@@ -9,7 +9,8 @@ const SnippetBlock = ({
   leadingSpaces = 2,
   ...props
 }) => {
-  const raw = children.split(/\r?\n/)
+  // I don't understand why this is needed…
+  const raw = children.toString().split(/\r?\n/)
 
   let code = ''
 
--- a/.nvmrc
+++ b/.nvmrc
@@ -1,1 +1,1 @@
-lts/Gallium
+lts/Hydrogen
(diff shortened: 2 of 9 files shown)


lukeg90:
Why not simply remove the dead links so the build doesn't fail?

lukeg90 (commented):
Why not simply remove the dead links so the build doesn't fail?

ChrisChinchilla:
@ChrisChinchilla Oh I see now. I'm a bit confused by the CI checks, since I'm still getting a dead link error when I try to build locally, but now I see that the PR checks are showing a different error.

Update: I re-ran the checks and now they're all passing except for the AWS preview 🤷

src/components/SnippetBlock/index.js
```diff
@@ -9,7 +9,8 @@ const SnippetBlock = ({
   leadingSpaces = 2,
   ...props
 }) => {
-  const raw = children.split(/\r?\n/)
+  // I don't understand why this is needed…
+  const raw = children.toString().split(/\r?\n/)
```
lukeg90, line 13: `split` is a string method and must be called on a string. React children can be strings, but more often they will be JSX, which I guess would have to be converted to a string before splitting it like this.

The mystery for me is how it was working before.

lukeg90 (commented):
Nice work! That was no trivial task 🙂

src/components/SnippetBlock/index.js
lukeg90: btw when copying and pasting code you don't understand, it's best to include a link to the source

lukeg90 (approved):

src/components/SnippetBlock/index.js
ChrisChinchilla: @lukeg90 I didn't copy and paste, this came from IDE errors, and so I cast it, but much like you I didn't understand why it worked before, so basically I had the same thought process as you but forgot to remove my thought process 😆

PR #297, opened 2024-01-31
Refactor custom component

ChrisChinchilla:
Fixes broken custom component after #293

### DIFF
--- a/src/components/LogoText/index.js
+++ b/src/components/LogoText/index.js
@@ -2,9 +2,9 @@ import React from 'react'
 import clsx from 'clsx'
 import styles from './custom.module.css'
 import ThemedImage from '@theme/ThemedImage'
-import ReactMarkdown from 'react-markdown'
 
 const LogoText = ({ children, width, linkTo, alt, srcLight, srcDark }) => {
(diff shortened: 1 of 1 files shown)


src/components/LogoText/index.js
```diff
@@ -19,7 +19,7 @@ const LogoText = ({ children, width, linkTo, alt, srcLight, srcDark }) => {
           />
         </a>
       </div>
-      <div className={clsx(styles.sideImageR)}><ReactMarkdown>{children}</ReactMarkdown></div>
+      <div className={clsx(styles.sideImageR)}>{children.props.children}</div>
```
weichweich, line 22: That removes the markdown component. `ReactMarkdown` so i guess that was the issue?

weichweich (approved):
LGTM

src/components/LogoText/index.js
ChrisChinchilla: Yeah, @weichweich, I think something changed, and it wasn't needed anymore. Instead, the value passed to the component was already a React component and could be rendered without anything else.

PR #300, opened 2024-02-05
fix: snippet component

rflechtner:
## fixes a TODO

The snippets component implementation was highly brittle and not really useable. This regex-based implementation should be much more capable.

## How to test:

Check out the different examples where the snippet component or JsTsSnippet are used.

## Checklist:

    * Link the PR/Ticket here

### DIFF
--- a/docs/develop/07_dApp/04_verifier.md
+++ b/docs/develop/07_dApp/04_verifier.md
@@ -5,7 +5,6 @@ title: Verifying a Credential
 
 import TsJsSnippet from '@site/src/components/TsJsSnippet';
 import TsJsBlock from '@site/src/components/TsJsBlock';
-import TsJsSnippet from '@site/src/components/TsJsSnippet';
 
 import EmailCtype from '!!raw-loader!@site/code_examples/sdk_examples/src/dapp/verifier/01_email_ctype.ts';
 import GenerateChallenge from '!!raw-loader!@site/code_examples/sdk_examples/src/dapp/verifier/02_generate_challenge.ts';
--- a/docs/develop/07_dApp/04_verifier.md
+++ b/docs/develop/07_dApp/04_verifier.md
@@ -4,6 +4,7 @@ title: Verifying a Credential
 ---
 
 import TsJsBlock from '@site/src/components/TsJsBlock';
+import TsJsSnippet from '@site/src/components/TsJsSnippet';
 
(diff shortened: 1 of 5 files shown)


src/components/SnippetBlock/index.js
```diff
@@ -4,60 +4,30 @@ import CodeBlock from '@theme/CodeBlock'
 const SnippetBlock = ({
   children,
   funcName = 'main',
-  funcEnd = '}',
-  snippets,
   leadingSpaces = 2,
+  dropHead = 0,
```
ntn-x2, line 8: Can you add a comment on what these two new parameters mean?

ntn-x2 (approved):
I guess you took the regex from somewhere online. In any case, I guess it does its job since the snippets look the same as the old ones. Cool improvement! I guess you can disregard my review in the other PR fixing the snippets since in this one everything is a `TsJsBlock` or snippet. There are few places where we use `CodeBlock` when we would probably have to use `TsJsBlock`, but that is out of scope for this PR. Maybe we can make a ticket for it?

rflechtner:
> I guess you took the regex from somewhere online.

I'm saddened that you're doubting my regex proficiency! Actually built that myself :) 

It's either this or counting brackets, which may be even more robust.

> In any case, I guess it does its job since the snippets look the same as the old ones. 

I also ran a bunch of snippets against it in the regex playground. 

> There are few places where we use `CodeBlock` when we would probably have to use `TsJsBlock`, but that is out of scope for this PR. Maybe we can make a ticket for it?

Sure, let's do it

ChrisChinchilla:
Thanks for the eyes @rflechtner !

src/components/SnippetBlock/index.js
rflechtner: will do, they let you drop leading or trailing lines (e.g., if you want to not show the return statement).

ChrisChinchilla (approved):

docs/develop/07_dApp/04_verifier.md
```diff
@@ -5,6 +5,7 @@ title: Verifying a Credential
 
 import TsJsSnippet from '@site/src/components/TsJsSnippet';
 import TsJsBlock from '@site/src/components/TsJsBlock';
+import TsJsSnippet from '@site/src/components/TsJsSnippet';
```
ChrisChinchilla:
```suggestion
```

PR #304, opened 2024-02-27
Update banner copy

ChrisChinchilla:
it seemed about time to update the banner advertising a nearly two year old release. Not sure if this is the best thing to put instead and I am open to suggestions…


Dudleyneedham (approved):
LGTM!

PR #305, opened 2024-03-01
Add versioning to docs

ChrisChinchilla:
## fixes KILTProtocol/ticket#3149

Adds a new version component to allow us to specify which version of the runtime or SDK particular features belong to.

There are a few caveats:

- Are these all the versions of projects?
- The actually version strings and how we use them is open for discussion
- The way the component looks is open for discussion
- I need to figure out how we bring these into pages that are pulled from elsewhere, such as the two DIP docs here

### DIFF
--- a/docs/concepts/07_dip/04_user_account_kilt.md
+++ b/docs/concepts/07_dip/04_user_account_kilt.md
@@ -3,9 +3,7 @@ id: dip-accounts-kilt
 title: Enabling DIP for user accounts on the KILT blockchain
 ---
 
-import { siteVariables } from '@site/src/versions'
-
-:::version-label {siteVariables.dipVersion}
+:::version-label[DIP]
 
 :::
 
--- a/docs/concepts/07_dip/04_user_account_kilt.md
+++ b/docs/concepts/07_dip/04_user_account_kilt.md
@@ -3,6 +3,12 @@ id: dip-accounts-kilt
 title: Enabling DIP for user accounts on the KILT blockchain
 ---
 
+import {siteVariables} from '@site/src/versions'
+
+:::version-label {siteVariables.dipVersion}
+
+:::
+
 For an account to take advantage of DIP it needs a decentralized identity (DID) and to create a transaction on the provider chain to generate a cross-chain identity commitment.
 
 For an account to be able to do this, a Dapp developer needs to build the functionality into their app for a user using the DIP SDK.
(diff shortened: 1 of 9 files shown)


src/versions.js
```diff
@@ -0,0 +1,4 @@
+export const siteVariables = {
+    dipVersion: 'Feature available in version 1.12.0 of the KILT runtime and higher',
```
rflechtner: Isn't DIP only available on the 1.12.0 version of peregrine, but not yet on spiritnet?

ChrisChinchilla:
@rflechtner and @Dudleyneedham I feel that this might need expanding in terms if figuring out what functionality matches what versions, but what are opinions on the implementation?

rflechtner:
> @rflechtner and @Dudleyneedham I feel that this might need expanding in terms if figuring out what functionality matches what versions, but what are opinions on the implementation?

Makes a lot of sense for incremental feature updates, but how we would set up the documentation for sdk v1 vs v0.xx and similar cases where the way certain things are done differ between versions is still unclear to me. For that we'd have to develop a vision too in the near future, given that the release of v1 is getting closer.

ChrisChinchilla:
@rflechtner I was trying to avoid setting up full versioning of docs, as while that's possible, I think it's more trouble than it's worth for the relatively small changes we have. Here's an idea… We could just use code tabs to highlight differences between supported versions? I'll throw an example into this PR

ChrisChinchilla:
@rflechtner Ready for a new look!

docs/concepts/07_dip/05_dapp_developer.md
```diff
@@ -6,6 +6,12 @@ title: Dapp developer
 import Tabs from '@theme/Tabs';
 import TabItem from '@theme/TabItem';
 
+import {siteVariables} from '@site/src/versions'
```
Dudleyneedham:
```suggestion
import { siteVariables } from '@site/src/versions'
```

docs/concepts/07_dip/04_user_account_kilt.md
```diff
@@ -3,6 +3,12 @@ id: dip-accounts-kilt
 title: Enabling DIP for user accounts on the KILT blockchain
 ---
 
+import {siteVariables} from '@site/src/versions'
```
Dudleyneedham:
```suggestion
import { siteVariables } from '@site/src/versions'
```

Dudleyneedham (approved):
LGTM. Minor changes

ChrisChinchilla:
OK, will merge for now, and we can add version warnings throughout docs as we find them

PR #306, opened 2024-03-21
Rework workshop to use new SDK methods

ChrisChinchilla:
## fixes KILTProtocol/ticket#3130

Changes workshop to use simpler `createFromAccount` method for DID creation.

### DIFF
--- a/code_examples/sdk_examples/src/dapp/dapp/04_attest_credential.ts
+++ b/code_examples/sdk_examples/src/dapp/dapp/04_attest_credential.ts
@@ -1,3 +1,4 @@
+/* eslint-disable @typescript-eslint/no-unused-vars */
 import * as Kilt from '@kiltprotocol/sdk-js'
 
 export async function main({
--- a/code_examples/sdk_examples/src/dapp/dapp/04_attest_credential.ts
+++ b/code_examples/sdk_examples/src/dapp/dapp/04_attest_credential.ts
@@ -22,15 +22,15 @@ export async function main({
   // We authorize the call using the attestation key of the Dapps DID.
   const extrinsic = api.tx.did.dispatchAs(dappAccount.address, attestationTx)
 
-  // const submitTx = await Kilt.Did.authorizeTx(
-  //   didUri,
-  //   attestationTx,
-  //   async ({ data }) => ({
-  //     signature: assertionMethodKey.sign(data),
-  //     keyType: assertionMethodKey.type
-  //   }),
-  //   dappAccount.address
-  // )
+  const submitTx = await Kilt.Did.authorizeTx(
+    didUri,
+    attestationTx,
+    async ({ data }) => ({
+      signature: assertionMethodKey.sign(data),
+      keyType: assertionMethodKey.type
+    }),
+    dappAccount.address
+  )
 
   // Since DIDs can not hold any balance, we pay for the transaction using our blockchain account
   const result = await Kilt.Blockchain.signAndSubmitTx(extrinsic, dappAccount)
--- a/code_examples/sdk_examples/src/dapp/dapp/04_attest_credential.ts
+++ b/code_examples/sdk_examples/src/dapp/dapp/04_attest_credential.ts
@@ -22,16 +22,6 @@ export async function main({
   // We authorize the call using the attestation key of the Dapps DID.
   const extrinsic = api.tx.did.dispatchAs(dappAccount.address, attestationTx)
 
-  const submitTx = await Kilt.Did.authorizeTx(
-    didUri,
-    attestationTx,
-    async ({ data }) => ({
-      signature: assertionMethodKey.sign(data),
-      keyType: assertionMethodKey.type
-    }),
-    dappAccount.address
-  )
-
   // Since DIDs can not hold any balance, we pay for the transaction using our blockchain account
   const result = await Kilt.Blockchain.signAndSubmitTx(extrinsic, dappAccount)
 
--- a/code_examples/sdk_examples/src/dapp/dapp/04_attest_credential.ts
+++ b/code_examples/sdk_examples/src/dapp/dapp/04_attest_credential.ts
@@ -12,7 +12,6 @@ export async function main({
   domainLinkageCredential: Kilt.ICredential
 }) {
   const api = Kilt.ConfigService.get('api')
-
   const { cTypeHash, claimHash } = Kilt.Attestation.fromCredentialAndDid(
     domainLinkageCredential,
     didUri
@@ -20,18 +19,20 @@ export async function main({
   const attestationTx = api.tx.attestation.add(claimHash, cTypeHash, null)
 
   // We authorize the call using the attestation key of the Dapps DID.
-  const submitTx = await Kilt.Did.authorizeTx(
-    didUri,
-    attestationTx,
-    async ({ data }) => ({
-      signature: assertionMethodKey.sign(data),
-      keyType: assertionMethodKey.type
-    }),
-    dappAccount.address
-  )
+  const extrinsic = api.tx.did.dispatchAs(dappAccount.address, attestationTx)
+
+  // const submitTx = await Kilt.Did.authorizeTx(
+  //   didUri,
+  //   attestationTx,
+  //   async ({ data }) => ({
+  //     signature: assertionMethodKey.sign(data),
+  //     keyType: assertionMethodKey.type
+  //   }),
+  //   dappAccount.address
+  // )
 
   // Since DIDs can not hold any balance, we pay for the transaction using our blockchain account
-  const result = await Kilt.Blockchain.signAndSubmitTx(submitTx, dappAccount)
+  const result = await Kilt.Blockchain.signAndSubmitTx(extrinsic, dappAccount)
 
   if (result.isError) {
     console.log('Attestation failed')
--- a/code_examples/sdk_examples/src/workshop/attester/attestCredential.ts
+++ b/code_examples/sdk_examples/src/workshop/attester/attestCredential.ts
@@ -11,6 +11,8 @@ export async function attestCredential(
   attesterAccount: Kilt.KiltKeyringPair,
   attesterDid: Kilt.DidUri,
   credential: Kilt.ICredential,
+  // TODO: Sort later
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
   signCallback: Kilt.SignExtrinsicCallback
 ): Promise<void> {
   const api = Kilt.ConfigService.get('api')
--- a/code_examples/sdk_examples/src/workshop/attester/attestCredential.ts
+++ b/code_examples/sdk_examples/src/workshop/attester/attestCredential.ts
@@ -1,4 +1,3 @@
-/* eslint-disable @typescript-eslint/no-unused-vars */
 import { config as envConfig } from 'dotenv'
 
 import * as Kilt from '@kiltprotocol/sdk-js'
@@ -34,6 +33,7 @@ export async function attestingFlow(
   claimerDid: Kilt.DidUri,
   attesterAccount: Kilt.KiltKeyringPair,
   attesterDid: Kilt.DidUri,
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
   signCallback: Kilt.SignExtrinsicCallback
 ): Promise<Kilt.ICredential> {
   // First the claimer.
--- a/code_examples/sdk_examples/src/workshop/attester/attestCredential.ts
+++ b/code_examples/sdk_examples/src/workshop/attester/attestCredential.ts
@@ -23,12 +23,7 @@ export async function attestCredential(
 
   // Create the tx and authorize it.
   const tx = api.tx.attestation.add(claimHash, cTypeHash, null)
-  const extrinsic = await Kilt.Did.authorizeTx(
-    attesterDid,
-    tx,
-    signCallback,
-    attesterAccount.address
-  )
+  const extrinsic = api.tx.did.dispatchAs(attesterAccount.address, tx)
 
   // Submit the tx to write the attestation to the chain.
   console.log('Attester -> create attestation...')
--- a/code_examples/sdk_examples/src/workshop/attester/attestCredential.ts
+++ b/code_examples/sdk_examples/src/workshop/attester/attestCredential.ts
@@ -44,7 +44,7 @@ export async function attestingFlow(
   // ... send the request to the attester
 
   // The attester checks the attributes and attests the provided credential.
-  await attestCredential(attesterAccount, attesterDid, credential, signCallback)
+  await attestCredential(attesterAccount, attesterDid, credential)
 
   // Return the generated credential.
   return credential
--- a/code_examples/sdk_examples/src/workshop/attester/attestCredential.ts
+++ b/code_examples/sdk_examples/src/workshop/attester/attestCredential.ts
@@ -1,3 +1,4 @@
+/* eslint-disable @typescript-eslint/no-unused-vars */
 import { config as envConfig } from 'dotenv'
 
 import * as Kilt from '@kiltprotocol/sdk-js'
@@ -10,7 +11,7 @@ import { generateLightDid } from '../claimer/generateLightDid'
 export async function attestCredential(
   attesterAccount: Kilt.KiltKeyringPair,
   attesterDid: Kilt.DidUri,
-  credential: Kilt.ICredential,
+  credential: Kilt.ICredential
 ): Promise<void> {
   const api = Kilt.ConfigService.get('api')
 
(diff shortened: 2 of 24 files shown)


ChrisChinchilla:
@rflechtner and @Ad96el A lot is still needed here from a reorganising perspective (it removes so much, which I will put elsewhere), but initially, are the `createFromAccount` changes correct?

code_examples/sdk_examples/src/workshop/attester/generateDid.ts
```diff
@@ -1,51 +1,40 @@
...
-      assertionMethod: [assertionMethod],
-      capabilityDelegation: [capabilityDelegation]
-    },
-    submitterAccount.address,
-    async ({ data }) => ({
-      signature: authentication.sign(data),
-      keyType: authentication.type
+export async function createFullDid() {
+  const api = await Kilt.connect(process.env.WSS_ADDRESS as string);
+  const accountMnemonic = process.env.ATTESTER_ACCOUNT_MNEMONIC as string;
```
rflechtner: in principle this works, but these files are tested, right? creating a DID can only ever be done once, so we'd have to create a new ATTESTER_ACCOUNT_MNEMONIC and send funds to the account before each run?

code_examples/sdk_examples/src/workshop/attester/generateDid.ts
```diff
@@ -1,51 +1,40 @@
 import { config as envConfig } from 'dotenv'
 
 import * as Kilt from '@kiltprotocol/sdk-js'
-
+// TODO: No longer needed in content, but needed for tutorial?
+// TODO: What to do?
+// TODO: And generateAccount needed changes, why?
 import { generateAccount } from './generateAccount'
```
rflechtner, line 3: can't we use the generated account to create a DID?

ChrisChinchilla: @rflechtner Yes, I guess I wanted to ensure it was "the right sort of account" to demonstrate the feature. I assume it is, my question here was also more around why we needed the changes here - https://github.com/KILTprotocol/docs/blob/1213e05ff9720775cac2c09fcfa246219ed71e72/code_examples/sdk_examples/src/workshop/attester/generateAccount.ts#L8

code_examples/sdk_examples/src/workshop/attester/generateDid.ts
ChrisChinchilla: @rflechtner Yes, but we always used to create a DID anyway, so actually I am not sure how the tests ever worked. Couple of options to balance getting our tests to work and a tutorial people can follow…

1. Create a new random DID every time
2. Create a new account each time, to create a new DID each time etc
3. Wrap the code to only create a DID if it doesn't already exist

I think 3 is the simplest and tidiest, and we make sure the tutorial is clear. You need to plug your own data in, etc, maybe through the _.env_ file.

code_examples/sdk_examples/src/workshop/attester/generateDid.ts
rflechtner: That, I think, was due to how the input for the api.tx.did.* functions below are typed. The `type` key must be one of a set of known values, and somehow it was typed as `string` unless we added that.

code_examples/sdk_examples/src/workshop/attester/generateDid.ts
rflechtner: 1. Is what we used to do before, but is not an option because we now create the did based on an account.
Tweaked to make the "random did" approach work again gives us 2. -> make random account, send accounts tokens, create did from account. I think that's the best solution because if we follow 3 everywhere, we're not actually testing much, we'll just skip through all creation operations

ChrisChinchilla: @rflechtner fair enough, i guess it's on test net anyway

ChrisChinchilla:
@rflechtner Ready for a re-review, the link checker is failing on an external link again that we can't control much…

code_examples/sdk_examples/src/workshop/attester/generateDid.ts
```diff
@@ -1,9 +1,6 @@
-import { config as envConfig } from 'dotenv'
-
 import * as Kilt from '@kiltprotocol/sdk-js'
-
+import { config as envConfig } from 'dotenv'
 import { generateAccount } from './generateAccount'
-import { generateKeypairs } from './generateKeypairs'
 
 export async function createFullDid(
   submitterAccount: Kilt.KiltKeyringPair
```
rflechtner: we could rename this parameter now to reflect that it's the account from which the DID is created, which controls which DID is created too.

code_examples/sdk_examples/src/workshop/attester/generateDid.ts
```diff
@@ -12,40 +9,33 @@ export async function createFullDid(
...
-  )
-
-  await Kilt.Blockchain.signAndSubmitTx(fullDidCreationTx, submitterAccount)
-
-  const didUri = Kilt.Did.getFullDidUriFromKey(authentication)
-  const encodedFullDid = await api.call.did.query(Kilt.Did.toChain(didUri))
-  const { document } = Kilt.Did.linkedInfoFromChain(encodedFullDid)
-
-  if (!document) {
+  const mnemonic = process.env.ATTESTER_ACCOUNT_MNEMONIC as string
```
rflechtner: This doesn't really make sense to me. The `mnemonic` used to be the one from which the DID is derived, but now it's the attester mnemonic - and the link between this mnemonic and the DID has become completely unclear.

code_examples/sdk_examples/src/workshop/attester/generateDid.ts
```diff
@@ -12,40 +9,33 @@ export async function createFullDid(
...
+  const { type, publicKey } = submitterAccount
+
+  const txs = [
+    api.tx.did.createFromAccount({ [type as 'ed25519']: publicKey }),
```
rflechtner: we could also get around this type forcing by changing the function parameter on line 6 to 
`submitterAccount: Kilt.KiltKeyringPair & { type: 'ed25519' }`

code_examples/sdk_examples/src/workshop/attester/generateDid.ts
```diff
@@ -12,40 +9,33 @@ export async function createFullDid(
...
+  console.log('Creating DID from account…')
+  const didDocument = await Kilt.Blockchain.signAndSubmitTx(
+    api.tx.utility.batch(txs),
+    submitterAccount
+  ).then(async () => {
+    const didUri = Kilt.Did.getFullDidUriFromKey(submitterAccount)
+    const encodedFullDid = await api.call.did.query(Kilt.Did.toChain(didUri))
+    const { document } = Kilt.Did.linkedInfoFromChain(encodedFullDid)
+    return document
+  })
```
rflechtner: I don't understand the use of a then clause here - we're in an async context and signAndSubmit is already awaited.
If you wanted to simply rename the variable you could do
```
 const { document: didDocument } = Kilt.Did.linkedInfoFromChain(encodedFullDid)
```
What was the goal for which you introduced it?

docs/develop/03_workshop/04_attester/02_did.md
```diff
@@ -117,15 +104,17 @@ Now run the code with:
   </TabItem>
 </Tabs>
 
-Once you have run the script, the output should provide you with your `ATTESTER_DID_MNEMONIC` and `ATTESTER_DID_URI`.
-The output should look like the following, but not identical since the DIDs are constructed from your account:
+Once you have run the script, the output should provide you with the `ATTESTER_DID_MNEMONIC` and `ATTESTER_DID_URI`.
+Using the `createFromAccount` method means that the `ATTESTER_DID_MNEMONIC` and `ATTESTER_ACCOUNT_MNEMONIC` are the same.
```
rflechtner: So it seems like the script now simply reads the supplied mnemonic back to the user, which I don't think that it makes a lot of sense to do. I'd say we best remove the mnemonic bit from `createFullDid`. And we could replace `ATTESTER_DID_MNEMONIC` with `ATTESTER_ACCOUNT_MNEMONIC` everywhere.

docusaurus.config.js
```diff
@@ -234,39 +234,40 @@ module.exports = {
...
+              documents: ["README.md"], // the file names to download,
+              modifyContent(filename, content) {
+                if (filename.includes("README")) {
+                  return {
+                    filename: "03_consumer.md",
+                    content: content
+                  }
+                }
+                return undefined
+              },
```
rflechtner: what changed here?

code_examples/sdk_examples/src/workshop/attester/attestCredential.ts
```diff
@@ -11,6 +11,8 @@ export async function attestCredential(
   attesterAccount: Kilt.KiltKeyringPair,
   attesterDid: Kilt.DidUri,
   credential: Kilt.ICredential,
+  // TODO: Sort later
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
   signCallback: Kilt.SignExtrinsicCallback
```
rflechtner: that is indeed not needed any more

rflechtner (commented):
Looks solid, I just think that where we now have only one mnemonic instead of two (because attester account and attester did are the same) we should consolidate both env variables

code_examples/sdk_examples/src/workshop/attester/generateDid.ts
ChrisChinchilla: @rflechtner Probably because I am also quite unclear what should be happening. using the new `createFromAccount`, can the DID account and the attester still have the same relationship as before?

docusaurus.config.js
ChrisChinchilla: @rflechtner I'm not sure, maybe linting…

code_examples/sdk_examples/src/workshop/attester/generateDid.ts
ChrisChinchilla: @rflechtner I thought that was "best practice", but if it's not necessary, I have simplified it.

rflechtner: in an async context it is almost always more legible if you simply await other async functions, I'd say

ChrisChinchilla:
@rflechtner Ready for a final review, I can already see a lot of next steps tweaks…

code_examples/sdk_examples/src/dapp/dapp/04_attest_credential.ts
```diff
@@ -12,26 +13,27 @@ export async function main({
...
+
+  // const submitTx = await Kilt.Did.authorizeTx(
+  //   didUri,
+  //   attestationTx,
+  //   async ({ data }) => ({
+  //     signature: assertionMethodKey.sign(data),
+  //     keyType: assertionMethodKey.type
+  //   }),
+  //   dappAccount.address
+  // )
```
rflechtner: we'd probably get rid of this before merging?

code_examples/sdk_examples/src/workshop/attester/generateDid.ts
```diff
@@ -59,10 +46,10 @@ if (require.main === module) {
       // Load attester account
       const accountMnemonic = process.env.ATTESTER_ACCOUNT_MNEMONIC as string
       const { account } = generateAccount(accountMnemonic)
-      const { mnemonic, fullDid } = await createFullDid(account)
+      const { fullDid } = await createFullDid(account)
 
       console.log('\nsave following to .env to continue\n')
-      console.error(`ATTESTER_DID_MNEMONIC="${mnemonic}"\n`)
+      // console.error(`ATTESTER_DID_MNEMONIC="${mnemonic}"\n`)
```
rflechtner: this can probably also go?

code_examples/sdk_examples/src/workshop/attester/generateDid.ts
```diff
@@ -1,51 +1,38 @@
...
-
 import * as Kilt from '@kiltprotocol/sdk-js'
-
+import { config as envConfig } from 'dotenv'
 import { generateAccount } from './generateAccount'
-import { generateKeypairs } from './generateKeypairs'
 
 export async function createFullDid(
-  submitterAccount: Kilt.KiltKeyringPair
+  creatorAccount: Kilt.KiltKeyringPair & { type: 'ed25519' }
```
rflechtner: The more correct (because more inclusive) typing would be the following, which we also use in a few other locations in the docs:
```suggestion
  creatorAccount: Kilt.KiltKeyringPair & { type: 'ed25519' | 'sr25519' | 'ecdsa' }
```

rflechtner: This is the complete set of key types that can be handled by the createFromAccount and setAttestationKey transactions.

docs/develop/03_workshop/04_attester/01_account.md
```diff
@@ -66,14 +66,9 @@ The `generateAccount` method returns an object with the following two properties
 Generating these values takes two steps:
 
 1. Create the `mnemonic` value using the `mnemonicGenerate()` method from the `Utils.Crypto` package.
-2. The `account` value first needs a `keyring` value defined, which is a data structure for defining the key pair type with the following parameters:
+2. The `account` value first needs a `keyring` value defined, which is a data structure for defining the key pair type as `ed25519`.
```
rflechtner: Ed25519 is simply one of the 3 key types/signature algorithms supported. We typed everything as `ed25519` because that was easiest (and is the default if you don't change the key type using the `type` parameter on `makeKeypairFromUri`), but strictly speaking any of 'sr25519', 'ed25519', or 'ecdsa' work.

docs/develop/03_workshop/04_attester/02_did.md
rflechtner: Do you think it still makes sense to mention the ATTESTER_DID_MNEMONIC at all? Should we not remove it from the docs and the code when it's identical to the ATTESTER_ACCOUNT_MNEMONIC ?

rflechtner (commented):
Makes sense to me! I think we should align the Keypair typing with other such typings throughout the docs (which use the union of all key pair types that can be used on a DID document as of now). But no major complaints here!

docs/develop/03_workshop/04_attester/02_did.md
ChrisChinchilla: Good call, will remove

code_examples/sdk_examples/src/dapp/dapp/04_attest_credential.ts
ChrisChinchilla: That was another good catch. Thought I'd caught them all…

ChrisChinchilla:
@rflechtner Some good catches, all resolved

code_examples/sdk_examples/src/dapp/dapp/04_attest_credential.ts
```diff
@@ -12,14 +13,15 @@ export async function main({
...
   const { cTypeHash, claimHash } = Kilt.Attestation.fromCredentialAndDid(
     domainLinkageCredential,
     didUri
   )
   const attestationTx = api.tx.attestation.add(claimHash, cTypeHash, null)
 
   // We authorize the call using the attestation key of the Dapps DID.
+  const extrinsic = api.tx.did.dispatchAs(dappAccount.address, attestationTx)
+
   const submitTx = await Kilt.Did.authorizeTx(
```
rflechtner: that bit is not in use any more now, am I right?

docs/develop/03_workshop/04_attester/01_account.md
```diff
@@ -66,14 +66,9 @@ The `generateAccount` method returns an object with the following two properties
 Generating these values takes two steps:
 
 1. Create the `mnemonic` value using the `mnemonicGenerate()` method from the `Utils.Crypto` package.
-2. The `account` value first needs a `keyring` value defined, which is a data structure for defining the key pair type with the following parameters:
+2. The `account` value first needs a `keyring` value defined, which is a data structure for defining the key pair type. This example uses `ed25519`, but `sr25519`, `ed25519`, or `ecdsa` are also valid.
```
rflechtner:
```suggestion
2. The `account` value first needs a `keyring` value defined, which is a data structure for defining the key pair type. This example uses `ed25519`, but `sr25519` or `ecdsa` are also valid.
```

rflechtner: or is there a reason to mention this twice?

rflechtner (approved):
We should probably tidy up dead code before we merge, but if you want to do it afterwards I'm also happy

code_examples/sdk_examples/src/dapp/dapp/04_attest_credential.ts
ChrisChinchilla: Looks like yes.

PR #308, opened 2024-04-04
Fix tip proposal doc

ChrisChinchilla:
There is a broken link on this page that was breaking tests. That page is now in the Polkadot archive and wasn't very helpful anyway, so I removed it and whilst I was there gave the page a general bit of polish.


rflechtner (approved):
LGTM!

PR #310, opened 2024-04-11
Fix broken DIP link

ChrisChinchilla:
I think out link checking needs some work and not completely sure what happened to this link.


ntn-x2 (commented):
Some tests fail, otherwise LGTM!

ChrisChinchilla:
@ntn-x2 Attempting to fix those elsewhere…

PR #322, opened 2024-07-25
SDK v1 Example

abdulmth:


examplesv1/src/createAndUpdate.ts
```diff
@@ -0,0 +1,220 @@
+import * as Kilt from '@kiltprotocol/sdk-js'
+import { Crypto } from '@kiltprotocol/utils'
+import { mnemonicGenerate, randomAsU8a } from '@polkadot/util-crypto'
+
+async function createDid() {
+
+  // Create a connection to a KILT blockchain.
+  // let api = await Kilt.connect('wss://peregrine.kilt.io/')
+  let api = await Kilt.connect('ws://localhost:9944/')
```
rflechtner, line 9: maybe just use a descriptive const?
```suggestion
  let api = await Kilt.connect(FULL_NODE_WEB_SOCKET_URL)
```

examplesv1/src/createAndUpdate.ts
```diff
@@ -0,0 +1,220 @@
...
+
+  // Create keypair, of which the DID will be derived.
+  const mnemonic = mnemonicGenerate()
+  const didKeypair = Crypto.makeKeypairFromUri(mnemonic, 'sr25519')
+
+  // The submitter account must have enough funds to cover the deposit costs.
+  const submitterMnemonic =
+    'build hill second flame trigger simple rigid cabbage phrase evolve final eight'
+  const submitter = Crypto.makeKeypairFromUri(submitterMnemonic, 'sr25519')
```
rflechtner, line 18: this will change in the next prerelease, so when you port this over to bundle tests, use `generateKeypair` instead.

examplesv1/src/createAndUpdate.ts
```diff
@@ -0,0 +1,220 @@
...
+    submitter,
+    fromPublicKey: didKeypair,
+  })
+
+  // The `createDid` function returns a transaction handler, which includes two methods:
+  // - `submit`: Submits a transaction for inclusion in a block, resulting in its execution in the blockchain runtime.
+  // - `getSubmittable`: Produces transaction that can be submitted to a blockchain node for inclusion, or signed and submitted by an external service.
+
+  // Submit transaction.
+  // Note: `submit()` by default, waits for the block to be finalized. This behaviour can be overwritten
```
rflechtner, line 47: I think the default currently is `inBlock` - that's something we should discuss again

abdulmth: but the `awaitFinalized` is true by default, no?

PR #327, opened 2024-08-20
SDK V1 docs

ChrisChinchilla:
Fixes https://github.com/KILTprotocol/ticket/issues/3576.


ChrisChinchilla:
@rflechtner Current status:

- Some of the versioning doesn't work as the build is broken, so it's commented out for now
- Most of the new tutorial is there, but code snippets need to be added
- I started customising `SnippetBlock` to accept a function name. It sort of works, but then it keeps going, so some work on the top and tail is needed.
- The verification step in the test code doesn't work
- Some more tear down methods are probably needed

PR #334, opened 2024-12-05
fix: pipeline

ntn-x2:


docusaurus.config.js
```diff
@@ -287,34 +287,5 @@ module.exports = {
         },
       },
     ],
-    [
```
Dudleyneedham, line 290: What was this?

ntn-x2: Yeah it was pulling in some READMEs from the chain repo and updating some links, but the final document looks exactly the same with and without it, so I just removed it.

Dudleyneedham: I should ask what was it intending to do

ntn-x2: Then my previous comment still answers your question, I think 🤔

Dudleyneedham (approved):
LGTM

PR #340, opened 2025-02-17
feat: tiny change to create a pr 3735

aybarsayan:
When I talked to Dudley, he said I needed to make my PRs smaller, so I'm starting from the very beginning. This is exactly the same as the current main, and if you approve this, I will make another small change and proceed from there. (Note: I was unable to send PR because its litterily the main branch, so I did the littlest change and then send a pr)

### DIFF
--- a/docusaurus.config.js
+++ b/docusaurus.config.js
@@ -259,7 +259,7 @@ module.exports = {
               content: trimContent,
             }
           }
-          return undefined
+          return undefined;
         },
       },
     ],


docusaurus.config.js
```diff
@@ -259,7 +259,7 @@ module.exports = {
               content: trimContent,
             }
           }
-          return undefined
+          return undefined;
```
Dudleyneedham, line 262: Lol. Is this to make a PR?

aybarsayan: Yes 😅 otherwise it doesn't allow me to create a pr so I did the simplest and littest change I possibly can 😅

rflechtner (approved):
So you're going to open PRs against the `new-sdkv1` branch now?

